PR #1, opened 2016-09-20
Allow defining and external location for the system installation and use it

mariospr:
This is needed as a first step to having apps installed both in the system
directory and an external location, even though for now is just one or another
but not both.

This commit defines the --with-external-install-dir configuration parameter
that allows specifying a path which is considered at runtime: if it exists,
that one will be used, otherwise the default system directory will be picked.

https://phabricator.endlessm.com/T12030

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -262,6 +262,14 @@ AC_ARG_WITH(system-install-dir,
 SYSTEM_INSTALL_DIR=$with_system_install_dir
 AC_SUBST(SYSTEM_INSTALL_DIR)
 
+AC_ARG_WITH(external-install-dir,
+           [AS_HELP_STRING([--with-external-install-dir=DIR],
+                           [Location of the external installation [LOCALSTATEDIR/endless-extra/flatpak]])],
+           [],
+           [with_external_install_dir='$(localstatedir)/endless-extra/flatpak'])
+EXTERNAL_INSTALL_DIR=$with_external_install_dir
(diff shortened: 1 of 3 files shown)


configure.ac
```diff
@@ -262,6 +262,14 @@ AC_ARG_WITH(system-install-dir,
 SYSTEM_INSTALL_DIR=$with_system_install_dir
 AC_SUBST(SYSTEM_INSTALL_DIR)
 
+AC_ARG_WITH(external-install-dir,
+           [AS_HELP_STRING([--with-external-install-dir=DIR],
+                           [Location of the external installation [LOCALSTATEDIR/endless-extra/flatpak]])],
+           [],
+           [with_external_install_dir='$(localstatedir)/endless-extra/flatpak'])
+EXTERNAL_INSTALL_DIR=$with_external_install_dir
+AC_SUBST(EXTERNAL_INSTALL_DIR)
```
dbnicholson: It would probably be nicer if this defaulted to not being set (`g_path_is_absolute` would catch the empty string case) rather than encoding the endless default here, but I'll let you wrangle that with upstream. I guess this makes it nicer for Endless developers hacking on flatpak.

mariospr: I think you're absolutely right. TBH, I've never consider this as an upstreameable thing due to its "short-term fix" nature, but it could actually be interesting upstream as an intermediate step too, so I'll keep that in mind. Thanks

PR #4, opened 2016-09-24
Temporary fix for knowledge applications in split disk layouts

mariospr:
The right solution would be to adapt the bundle manifests for each app,
but that would require rebundling all of them, but that's not feasible
for the incoming release so, as a temporary solution, we simply add an
extra default override in flatpak for each application, so that they
can automatically read /var/endless-extra/flatpak/app/APP_ID it they
are running in a split disk layout.

Note: this commit can safely be reverted once all the bundles have been
adapted to include the path for split disk layouts.

https://phabricator.endlessm.com/T13509

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2566,6 +2566,19 @@ compute_permissions (GKeyFile *app_metadata,
 
   add_default_permissions (app_context);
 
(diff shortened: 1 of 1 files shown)


mariospr:
We are going for the proper fix in the end, so this hack is (fortunately) no longer needed. Closing the PR

PR #8, opened 2016-10-17
flatpak-utils: accommodate old search provider names

mattdangerw:
I'm not convinced this should get merged, but tossing it out as an idea...

https://phabricator.endlessm.com/T13175


mattdangerw:
Wrong branch...

PR #9, opened 2016-10-17
flatpak-utils: accommodate old search provider names

mattdangerw:
https://phabricator.endlessm.com/T13735


mattdangerw:
Not convinced this should be merged, opening the PR to discuss.

PR #17, opened 2017-01-03
dir: make it possible to disable triggers sandbox with an env var

cosimoc:
This will be useful when an automated builder runs the triggers
chrooted.

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2316,10 +2316,19 @@ flatpak_dir_run_triggers (FlatpakDir   *self,
   g_autoptr(GFile) triggersdir = NULL;
   GError *temp_error = NULL;
   const char *triggerspath;
+  const char *disable_sandboxed_env;
+  gboolean disable_sandboxed_triggers = FALSE;
 
   triggerspath = g_getenv ("FLATPAK_TRIGGERSDIR");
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -2356,25 +2365,28 @@ flatpak_dir_run_triggers (FlatpakDir   *self,
...
-          g_ptr_array_add (argv_array, g_strdup ("--dev"));
-          g_ptr_array_add (argv_array, g_strdup ("/dev"));
-          g_ptr_array_add (argv_array, g_strdup ("--bind"));
-          g_ptr_array_add (argv_array, g_strdup (basedir));
-          g_ptr_array_add (argv_array, g_strdup (basedir));
-#endif
+          if (disable_sandboxed_triggers)
+            {
+              g_ptr_array_add (argv_array, g_file_get_path (child));
+              g_ptr_array_add (argv_array, g_strdup (basedir));
```
dbnicholson: I feel like this branch is not needed (and wasn't before in the `#ifdef` case). The `g_file_get_path (child)` and `g_strdup (basedir)` are repeated after the conditional. I think the way you have this, it would run `trigger path trigger path`. That said, all the triggers ignore any arguments after the 1st, so it probably doesn't matter in practice.

PR #18, opened 2017-01-04
Fix downloading extra data in libflatpak

joaquimrocha:
The download was getting stuck when launched from within a thread as is
the case when using GNOME Software.

This patch simplifies the code that downloads the extra data and drops
the use of a mainloop, calling the libsoup related code synchronously
instead. It also now returns an error when an issue occurred while the
data was being downloaded.

https://phabricator.endlessm.com/T14824


mariospr:
I wonder if this problem could be fixed by simply specifying its own GMainContext to the GMainLoop that's being created (so that running async ops from a different thread won't block)

PR #22, opened 2017-01-19
Added support for Mali drivers to --device=dri

xjuan:
https://phabricator.endlessm.com/T15102


ptomato:
This was committed upstream already.

PR #27, opened 2017-02-13
Don't user user namespaces when bubblewrap is setuid

mariospr:
It turns out that it is impossible for to get ptrace capabilities
for child user namespaces in the current kernel if the user
namespace is created as root, which is what happens when bwrap
is setuid root (see https://github.com/flatpak/flatpak/issues/557
for details).

This is very problematic, as ptrace rights controls access to
/proc/$pid/root which is what we base the detection of peer
app id and rights on for portals.

For now, we disable user namespaces (except for the case of
unprivileged user namespaces, where it is necessary and works).

https://phabricator.endlessm.com/T15430

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -3856,7 +3856,6 @@ flatpak_run_setup_base_argv (GPtrArray      *argv_array,
 
   add_args (argv_array,
             "--unshare-pid",
-            "--unshare-user-try",
             "--proc", "/proc",
             "--dir", "/tmp",
             "--dir", "/var/tmp",


erikos:
Testing feedback: the CodeView overlay button https://phabricator.endlessm.com/T15430 and the issue with the Processing application icon in the taskbar https://phabricator.endlessm.com/T15039 are fixed with this change.

common/flatpak-run.c
```diff
@@ -3856,7 +3856,6 @@ flatpak_run_setup_base_argv (GPtrArray      *argv_array,
 
   add_args (argv_array,
             "--unshare-pid",
-            "--unshare-user-try",
```
dbnicholson: I thought that the way `--unshare-user-try` worked (at least a while ago) was to not try user namespaces when bwrap is setuid. Wouldn't it be better to do such a fix in bubblewrap? (check if `getuid() != geteuid()` and ignore requests for user namespaces in that case)

BTW, Alex's blog suggests not using user namespaces when bwrap is setuid - https://blogs.gnome.org/alexl/2017/01/18/the-flatpak-security-model-part-1-the-basics/.

cosimoc:
Merging this, since it was already pushed upstream.
@dbnicholson I think the patch here implements your suggestion (not to use user namespaces with setuid bubblewrap), but feel free to comment on the upstream story/commit if required.

mariospr:
I just commented in the issue upstream to reflect Dan's point there: https://github.com/flatpak/flatpak/issues/557#issuecomment-279671949

PR #30, opened 2017-04-04
Whitelist apps with extra-data

GeorgesStavracas:
As a temporary workaround for the progress report
issue, add a whitelist of apps that has extra-data
and use that to detect extra-data apps.

This is meant to be a temporary downstream solution
until a proper fix lands upstream.

https://phabricator.endlessm.com/T15603

### DIFF
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -1121,6 +1121,25 @@ get_write_progress (guint outstanding_writes)
   return (guint) (3 / (gdouble) outstanding_writes);
 }
 
+static gboolean
+app_contains_extra_data (const char *app_id)
+{
+  /* HACK: This is a temporary solution until the detached metadata work is ready */
+  const char *extra_data_apps[] =
+    {
+      "com.dropbox.Client",
+      "com.google.Chrome",
+      "com.microsoft.Skype",
+      "org.mozilla.Firefox",
+      "com.slack.Slack",
+      "com.spotify.Client",
+      "org.videolan.VLC",
+      NULL
+    };
+
+  return app_id ? g_strv_contains (extra_data_apps, app_id) : FALSE;
+}
+
 static void
 progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
 {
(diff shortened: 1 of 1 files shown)


lib/flatpak-installation.c
```diff
@@ -1145,6 +1170,8 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
   guint64 current_time;
   guint new_progress = 0;
   gboolean estimating = FALSE;
+  gboolean has_extra_data;
+  char *app_id;
```
cosimoc: Should be `const char*`

lib/flatpak-installation.c
```diff
@@ -1121,6 +1121,31 @@ get_write_progress (guint outstanding_writes)
   return (guint) (3 / (gdouble) outstanding_writes);
 }
 
+static gboolean
+app_contains_extra_data (const char *app_id)
+{
+  /* HACK: This is a temporary solution until the detached metadata work is ready */
+  const char *extra_data_apps[] =
```
cosimoc: Do you need to hardcode the list here? Do you not have a way from inside flatpak to determine that the app had extra data?

GeorgesStavracas: Flatpak uses OSTree metadata to store the extra-data. Thus, it is only able to detect extra-data **after** the entire OSTree repo is pulled. Since we can't rely on that **while** the pull is happening, Flatpak can't detect it it has extra-data, not until it's too late.

Because we can't detect that while pulling, the pull progress goes up to 100% - then, when it finally detects extra-data, the progress already can't be rolled back.

If Flatpak used OSTree's detached metadata this wouldn't happen. I'm working on it as I write this comment, but still didn't managed to get something working. For 3.1.4, @rshuler proposed using this approach temporarily until the detached metadata solution is ready (for 3.2).

dbnicholson: This is just the way flatpak handles it. You can do a metadata only pull and examine it. We do that in eos-updater, but more because it predated the summary file. Speaking of which, we can always have flatpak include this information in the summary file, which I'm sure gnome-software is already relying on.

Not saying this won't work as is, but there are certainly ways to get this information.

PR #38, opened 2017-06-07
dir: make sure to whitelist our additional exports

cosimoc:
@erikos Please review


erikos:
Excellent looks good and should be what is needed to export the files fine on the OS itself.

PR #40, opened 2017-06-28
builder: Use build-args during cleanup

ptomato:
Any build-args specified in the manifest should be used during the
cleanup and platform-cleanup stages. This is because if you are using
QEMU to build for another architecture, for example, you need to pass
--bind-mount in the build-args, and the bind mount also needs to be
present while running cleanup commands.

https://phabricator.endlessm.com/T17867


ptomato:
This has been merged upstream: https://github.com/flatpak/flatpak/pull/877

pwithnall:
> This has been merged upstream: flatpak/flatpak#877

And it made it into the 0.9.7 release; we currently package flatpak 0.9.8 in OBS, so I think this PR can be closed and deleted. @ptomato?

PR #43, opened 2017-07-11
allow remote add to succeed when offline

ramcq:
Updating the metadata at the end of adding a remote means that it fails
if we're offline, even if we have the .flatpakrepo file available.
Unlike the other error cases in the process, which mean the operation
really has messed up, the remote is actually usable even if the
metadata update fails, because future operations will grab the
summary and any new metadata - so we should change this to a warning
to allow the operation to succeed when we're offline.

https://github.com/flatpak/flatpak/issues/917

For Endless, this should fix our repo adds appearing to fail when the
network isn't available at boot time.

https://phabricator.endlessm.com/T17172

### DIFF
--- a/app/flatpak-builtins-add-remote.c
+++ b/app/flatpak-builtins-add-remote.c
@@ -290,6 +290,7 @@ flatpak_builtin_add_remote (int argc, char **argv,
   g_autoptr(GKeyFile) config = NULL;
   g_autoptr(GBytes) gpg_data = NULL;
   gboolean changed = FALSE;
+  g_autoptr(GError) local_error = NULL;
 
   context = g_option_context_new (_("NAME LOCATION - Add a remote repository"));
   g_option_context_set_translation_domain (context, GETTEXT_PACKAGE);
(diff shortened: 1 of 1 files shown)


app/flatpak-builtins-add-remote.c
```diff
@@ -361,8 +362,13 @@ flatpak_builtin_add_remote (int argc, char **argv,
     return FALSE;
 
   /* We can't retrieve the extra metadata until the remote has been added locally, since
-     ostree_repo_remote_fetch_summary() works with the repository's name, not its URL. */
-  return flatpak_dir_update_remote_configuration (dir, remote_name, cancellable, error);
+     ostree_repo_remote_fetch_summary() works with the repository's name, not its URL.
+     Don't propagate errors here because the remote should be usable already. We
+     might just be offline. */
+  if (!flatpak_dir_update_remote_configuration (dir, remote_name, cancellable, &local_error))
+    g_printerr (_("Warning: could not update extra metadata for '%s': %s\n"), remote_name, local_error->message);
```
cosimoc: Instead of `g_printerr()` you can just use `g_warning()` (and remove `Warning:` from the message itself). Note that `g_warning()` also adds a newline.

cosimoc: Also, I wonder if you couldn't detect the specific "we are offline" error code from here; if you did, I'd simply ignore that error (or perhaps use `g_debug()` instead) and still propagate the other errors.

ramcq: Fixed, switched to g_warning and had a dig around - looks like G_IO_ERROR_FAILED is the best we can hope to get from ostree-fetcher-*/ostree-pull. It's not precise enough to not print a warning I don't think?

app/flatpak-builtins-add-remote.c
```diff
@@ -361,8 +362,23 @@ flatpak_builtin_add_remote (int argc, char **argv,
...
-     ostree_repo_remote_fetch_summary() works with the repository's name, not its URL. */
-  return flatpak_dir_update_remote_configuration (dir, remote_name, cancellable, error);
+     ostree_repo_remote_fetch_summary() works with the repository's name, not its URL.
+     Don't propagate IO failed errors here because we might just be offline - the
+     remote should already be usable. */
+  if (!flatpak_dir_update_remote_configuration (dir, remote_name, cancellable, &local_error))
+    {
+      if (g_error_matches (local_error, G_IO_ERROR, G_IO_ERROR_FAILED))
+        {
+          g_warning (_("Could not update extra metadata for '%s': %s\n"), remote_name, local_error->message);
```
cosimoc: It's very unusual to mark warnings for translation, and no other warnings are marked as such in the flatpak code base at the moment. Please change it to use a simple string.

ramcq: Hmm - the exact same string is used below so it's already translated. And all of the other strings printed out to the user in this file are translated, because they form part of the "UI" of these commands. I'm only using g_warning because you told me to... :P

ramcq: (actually almost none of the flatpak builtin commands use g_log that much, they just print stuff to stdout/stderr typically, with translated messages - that's why I used g_printerr in the first place)

PR #49, opened 2017-07-21
flatpak_get_gl_drivers() add support for mali

xjuan:
Look for mali driver version in /sys/module/mali/version which
contains driverversion_siliconversion and infer GL extension
named siliconversion-driverversion

https://phabricator.endlessm.com/T18288

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,6 +40,7 @@ FLATPAK_BINDIR=$(bindir)
 AM_CPPFLAGS =							\
 	-DFLATPAK_BINDIR=\"$(FLATPAK_BINDIR)\"			\
 	-DFLATPAK_SYSTEMDIR=\"$(SYSTEM_INSTALL_DIR)\"		\
+	-DFLATPAK_EXTERNALDIR=\"$(EXTERNAL_INSTALL_DIR)\"	\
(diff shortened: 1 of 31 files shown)


xjuan:
If you run this command an arm you should get this kind of output

```$ ./flatpak --gl-drivers
mali450r0p0-r6p0-01rel0
default
host
```

dsd:
How hard would it be just to get it to use the `host` driver? If that is what I assume, it seems like a good answer to the question of "how to make sure the runtime GL version matches the one in the host OS"

xjuan:
From https://lists.freedesktop.org/archives/flatpak/2017-February/000534.html
```
The current implementation of detecting the driver only has special
code for the nvidia binary driver, where it looks up
/sys/module/nvidia/version and generates a driver name like
nvidia-375-26. It also always adds the names "default" and "host" to
the list, which is supposed to be used for updating the default mesa
driver and for re-packaging the host driver as a flatpak driver.
Alternatively if the FLATPAK_GL_DRIVERS env var is set this overrides
the driver used.
```
I assume by "re-packaging" he means you could create an extension named
org.freedesktop.Platform.GL.host with the host drivers in it
but that would not work for us if we want to have two different arm extensions

PR #50, opened 2017-08-24
Update to upstream 0.9.8 release and enable --enable-p2p (T18242)

pwithnall:
https://phabricator.endlessm.com/T18242

PR #51, opened 2017-08-24
debian: Update to experimental 0.9.7-1 release, then 0.9.8, and enable --enable-p2p (T18242)

pwithnall:
https://phabricator.endlessm.com/T18242


pwithnall:
(Code changes are in PR #50.)

debian/control
```diff
@@ -94,18 +96,18 @@ Section: devel
...
  ${python3:Depends},
  ${shlibs:Depends},
 Recommends:
  binutils,
+ bzr,
  elfutils,
  git,
  patch,
  unzip,
-Suggests: bzr
```
dbnicholson: This had been intentionally changed by us downstream so bzr didn't get pulled into the image - https://github.com/endlessm/flatpak/commit/0694288131e3cf269abc8305b5313eaff9e93518.

debian/rules
```diff
@@ -7,19 +7,24 @@ export DEB_LDFLAGS_MAINT_APPEND = -Wl,--as-needed
...
 		--enable-docbook-docs \
 		--enable-gtk-doc \
 		--enable-installed-tests \
+		--enable-p2p \
 		--libexecdir=/usr/lib/flatpak \
 		--with-priv-mode=none \
 		--with-html-dir=/usr/share/doc/libflatpak-doc \
 		--with-privileged-group=sudo \
-		--with-system-bubblewrap=/usr/bin/bwrap \
+		--with-system-bubblewrap=bwrap \
```
dbnicholson: Another downstream change by us - https://github.com/endlessm/flatpak/commit/b197ff9b40d94c6a07dc19d168566344c78f682a. I'm not sure if this is still valid or not.

debian/flatpak.postinst
```diff
@@ -0,0 +1,15 @@
...
+
+if [ "$1" = configure ]; then
+    # Run a do-nothing command (it just lists configured remotes) for
+    # its side-effect of initializing the shared system-wide repository.
+    flatpak remote-list --system >/dev/null || :
+fi
+
+exit 0
+
+# vim:set sw=4 sts=4 et:
```
dbnicholson: We had intentionally removed this before - https://github.com/endlessm/flatpak/commit/b4429508650cafbdc49c4e93f33a889c41e62a13. However, I _think_ it's fine to bring it back. It will create /var/lib/flatpak in the ostree, but that gets masked at runtime by the OS /var, which the image builder (and tmpfiles.d) sets up with the proper symlinks.

debian/patches/series
```diff
@@ -1,2 +1 @@
-installed-tests-Install-test-keyring2-to-the-right-place.patch
-testlibrary-Call-g_assert_no_error-first.patch
+tests-Isolate-tests-from-real-home-directory-more-thoroug.patch
```
dbnicholson: You didn't actually add this patch, but I guess it doesn't matter since it gets dropped in 0.9.8.

debian/rules
mwleeds: This was fixed in the upstream commit 983b2a5db

debian/flatpak.postinst
mwleeds: I'll verify that once a nightly image is built with this flatpak.

debian/control
mwleeds: I reverted this and force pushed

debian/rules
dbnicholson: Ah, OK.

debian/control
dbnicholson: OK

debian/changelog
```diff
@@ -1,3 +1,57 @@
+flatpak (0.9.8-1endless1) experimental; urgency=medium
```
dbnicholson: Can you call this `0endless1`? Then if Debian catches up and we want to import their changes we can actually go forward in version.

mwleeds: good point

debian/control
pwithnall: Nitpick: It would be good to list this kind of deviation from the upstream Debian packaging in the `debian/changelog` in future, under the ‘resync to Debian packaging’ entry, to make it more obvious that it’s a diff we need to continue carrying.

PR #52, opened 2017-09-11
dir: relax requirements around xa.ref matching deployed commit

cosimoc:
In case the summary file is GPG-signed, we already have other better
ways of verifying expectations around the installed ref.
If the deployed branch does not match xa.ref, simply warn instead of
aborting.

Based on a patch by Dan Nicholson <nicholson@endlessm.com>

https://phabricator.endlessm.com/T17772


pwithnall:
It’s been proposed upstream as https://github.com/flatpak/flatpak/pull/1013.

PR #53, opened 2017-09-27
dir: Don't download GL nvidia extensions if "host" extension is installed

jprvita:
https://phabricator.endlessm.com/T19169

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -9494,6 +9494,32 @@ flatpak_dir_find_remote_related (FlatpakDir *self,
                   continue;
                 }
 
+              /* Don't download anything for "download-if=active-gl-driver"
+               * extension points if there is an unmanaged extension for it
+               * installed already */
+              if (g_strcmp0 (download_if, "active-gl-driver") == 0)
+                {
+                  const char **gl_drivers = flatpak_get_gl_drivers ();
+                  gboolean skip = FALSE;
+                  int d;
+
+                  for (d = 0; gl_drivers[d] != NULL; d++)
+                    {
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -9494,6 +9494,32 @@ flatpak_dir_find_remote_related (FlatpakDir *self,
...
+                {
+                  const char **gl_drivers = flatpak_get_gl_drivers ();
+                  gboolean skip = FALSE;
+                  int d;
+
+                  for (d = 0; gl_drivers[d] != NULL; d++)
+                    {
+                      g_autofree gchar *e = g_strconcat (extension, ".",
+                                                         gl_drivers[d], NULL);
+                      if (flatpak_find_unmaintained_extension_dir_if_exists(e, NULL, NULL, NULL))
```
ramcq: I think you want to set arch and branch to match...?

whitespace (

common/flatpak-dir.c
```diff
@@ -9494,6 +9494,32 @@ flatpak_dir_find_remote_related (FlatpakDir *self,
                   continue;
                 }
 
+              /* Don't download anything for "download-if=active-gl-driver"
+               * extension points if there is an unmanaged extension for it
+               * installed already */
```
ramcq: I think it doesn't necessarily make sense to have this logic be specific to unmanaged extensions. If our intended behavior is that, for example, the highest priority available driver is the best, it should be the same whether that extension is unmanaged or a regular one.

The only special-casing of the host extension in that case is that it would(/could/should/...) appear first in the drivers list, but because it's never available for download, stops any others being retrieved if it's available. I guess we'd also need to put it first in the list. :) But that doesn't seem too controversial to me.

PR #59, opened 2017-11-08
ls-remote: Fix crash when ostree-metadata ref exists in remote

pwithnall:
The code assumes that *all* refs in a remote contain a slash, which
technically is not the case, and pragmatically is definitely not the
case once the remote has an ostree-metadata ref.

Rebase 0.9.8.8: Affected code block has moved slightly from upstream.

https://phabricator.endlessm.com/T19867


dbnicholson:
Nothing yet upstream, but this is obviously correct. Going ahead with the merge.

dbnicholson:
Nothing yet upstream, but this is obviously correct. Going ahead with the merge.

PR #71, opened 2018-02-05
T21056

joaquimrocha:
https://phabricator.endlessm.com/T21056


joaquimrocha:
We have no notion of collection IDs in GNOME Software. I know my patch is treating the ref name as global and that loses precision, but that function already exists, so I was treating it as a "best match" approach, which I think is better than returning nothing.
Let's discuss this more in detail tomorrow.

PR #72, opened 2018-02-10
Rebase on 0.10.2 - DON'T MERGE

ptomato:
This is just for comparison to show the distro patches on top of 0.10.2.

From 0.10.1 to 0.10.2 it was a straightforward rebase with only a few merge conflicts with obvious solutions. I also dropped the revert of disabling app search by default, and the revert of that revert.

https://phabricator.endlessm.com/T21140


ptomato:
I'm going to close this for now, as I don't have enough time to spend on following through with the necessary testing. Also so that it doesn't get merged accidentally. I'll leave it up here for when the time comes to pick it up again.

PR #75, opened 2018-03-06
run: Add option to disallow a11y bus proxying

smspillaz:
This isn't needed for servers and starting the a11y bus on a
fresh session bus takes upwards of 15 seconds.

https://phabricator.endlessm.com/T21429

### DIFF
--- a/app/flatpak-builtins-run.c
+++ b/app/flatpak-builtins-run.c
@@ -44,6 +44,7 @@ static gboolean opt_devel;
 static gboolean opt_log_session_bus;
 static gboolean opt_log_system_bus;
 static gboolean opt_log_a11y_bus;
+static gboolean opt_no_a11y_bus;
 static gboolean opt_file_forwarding;
 static char *opt_runtime;
 static char *opt_runtime_version;
@@ -58,6 +59,7 @@ static GOptionEntry options[] = {
   { "log-session-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_session_bus, N_("Log session bus calls"), NULL },
   { "log-system-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_system_bus, N_("Log system bus calls"), NULL },
   { "log-a11y-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_a11y_bus, N_("Log accessibility bus calls"), NULL },
+  { "no-a11y-bus", 0, 0, G_OPTION_ARG_NONE, &opt_no_a11y_bus, N_("Don't proxy accessibility bus calls"), NULL },
   { "file-forwarding", 0, 0, G_OPTION_ARG_NONE, &opt_file_forwarding, N_("Enable file forwarding"), NULL },
   { NULL }
 };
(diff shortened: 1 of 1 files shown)


smspillaz:
Not sure if we want to upstream this - @cosimoc @pwithnall let me know

smspillaz:
> Why does Flatpak end up starting them if they are not running? Can't we fix that or detect the situation they are not in use by comparing ListNames to ListAvailableNames?

Not sure I understand your comment here. I think flatpak has always started all the a11y stuff if it is not running, probably by calling `GetAddress` and friends on `org.a11y.Bus` so that it can set up the proxies for them: https://github.com/endlessm/flatpak/blob/master/common/flatpak-run.c#L3212 . That will start them through dbus-activation.

In any event, when running a server, a11y doesn't matter so there's no point making all those calls for proxying.

pwithnall:
I think Rob’s point is: why should we introduce an option, and require server flatpak authors to think about using it, when we could get the same results with no option by making the a11y stuff start lazily?

smspillaz:
Both @ramcq and @pwithnall make a good point - the ideal here is to autostart the daemons when needed. I'm not sure if the current architecture permits that, since it seems like the AT_SPI bus address gets injected in at process-spawning time.

I guess my approach here is just in recognition that this is already an internal option. We could put it under the `--no-desktop` umbrella too.

app/flatpak-builtins-run.c
```diff
@@ -57,7 +58,8 @@ static GOptionEntry options[] = {
   { "runtime-version", 0, 0, G_OPTION_ARG_STRING, &opt_runtime_version, N_("Runtime version to use"), N_("VERSION") },
   { "log-session-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_session_bus, N_("Log session bus calls"), NULL },
   { "log-system-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_system_bus, N_("Log system bus calls"), NULL },
-  { "log-a11y-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_a11y_bus, N_("Log accessibility bus calls"), NULL },
+  { "log-a11y-bus", 0, 0, G_OPTION_ARG_NONE, &opt_log_a11y_bus, N_("Don't proxy accessibility bus calls"), NULL },
+  { "no-a11y-bus", 0, 0, G_OPTION_ARG_NONE, &opt_no_a11y_bus, N_("Log accessibility bus calls"), NULL },
```
smspillaz: Ugh, I also messed the ordering of these two strings :(

ramcq:
Oh! --no-desktop is a great idea. Add it there. :)

smspillaz:
> Oh! --no-desktop is a great idea. Add it there. :)

Damn, I just checked and it seems like `--no-desktop` was actually deprecated in b53b78b2a259abda529f2d9c756fa38ddab2c2de

app/flatpak-builtins-run.c
cosimoc: The order is still wrong.

smspillaz: must have forgotten to push my changes :)

smspillaz:
Note, since https://github.com/endlessm/eos-companion-app-integration/pull/72/commits/73ecb2e01a0b987cf2c3ee4da81435c7cb56e42f we can get away without requiring this change. Should I close this PR?

ramcq:
No, open it upstream. They might just merge it and we can drop the hack. We don't need that A11Y stuff running, and we're wasting heaps of RAM all over the OS, so let's try and fix this properly.

cosimoc:
Yes, let's close this downstream PR. Once/if it gets accepted upstream, we can reopen.

PR #79, opened 2018-04-11
dir: Make find_remote_ref(s) work offline

mwleeds:
This adds a function to fetch the xa.cache data from the ostree-metadata
ref, flatpak_dir_fetch_refs_cache, and uses it in
flatpak_dir_find_remote_ref and flatpak_dir_find_remote_refs. This makes
both functions work while offline because the ostree-metadata can be
fetched from peer sources such as USB drives or peers on a LAN (peer
sources don't redistribute remote summary files so those can't be used).

Fixes https://github.com/flatpak/flatpak/issues/1522

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7935,6 +7935,61 @@ typedef enum {
   FIND_MATCHING_REFS_FLAGS_KEEP_REMOTE = (1 << 0),
 } FindMatchingRefsFlags;
 
+/* helper function for find_matching_refs and find_matching_refs_from_cache */
+static void
+add_ref_if_match (FlatpakCollectionRef  *coll_ref,
+                  const char            *opt_name,
+                  const char            *opt_branch,
+                  const char            *opt_arch,
+                  const char            *opt_collection_id,
+                  FlatpakKinds           kinds,
+                  FindMatchingRefsFlags  flags,
+                  GPtrArray             *matched_refs)
+{
+  g_auto(GStrv) parts = NULL;
+  gboolean is_app, is_runtime;
+  const char **arches = flatpak_get_arches ();
+  const char *opt_arches[] = {opt_arch, NULL};
+  g_autofree char *ref = NULL;
+
+  if (opt_arch != NULL)
+    arches = opt_arches;
+
+  /* Unprefix any remote name if needed */
+  ostree_parse_refspec (coll_ref->ref_name, NULL, &ref, NULL);
+  if (ref == NULL)
+    return;
+
+  is_app = g_str_has_prefix (ref, "app/");
+  is_runtime = g_str_has_prefix (ref, "runtime/");
+
+  if ((!(kinds & FLATPAK_KINDS_APP) && is_app) ||
+      (!(kinds & FLATPAK_KINDS_RUNTIME) && is_runtime) ||
+      (!is_app && !is_runtime))
+    return;
+
+  parts = flatpak_decompose_ref (ref, NULL);
+  if (parts == NULL)
+    return;
+
+  if (opt_name != NULL && g_strcmp0 (opt_name, parts[1]) != 0)
+    return;
+
+  if (!g_strv_contains (arches, parts[2]))
+    return;
+
+  if (opt_branch != NULL && g_strcmp0 (opt_branch, parts[3]) != 0)
+    return;
+
+  if (opt_collection_id != NULL && g_strcmp0 (opt_collection_id, coll_ref->collection_id))
+    return;
+
+  if (flags & FIND_MATCHING_REFS_FLAGS_KEEP_REMOTE)
+    g_ptr_array_add (matched_refs, g_strdup (coll_ref->ref_name));
+  else
+    g_ptr_array_add (matched_refs, g_steal_pointer (&ref));
+}
+
 /* Guarantees to return refs which are decomposable. */
 static GPtrArray *
 find_matching_refs (GHashTable *refs,
@@ -7947,14 +8002,9 @@ find_matching_refs (GHashTable *refs,
                     GError      **error)
 {
   g_autoptr(GPtrArray) matched_refs = NULL;
-  const char **arches = flatpak_get_arches ();
-  const char *opt_arches[] = {opt_arch, NULL};
   GHashTableIter hash_iter;
   gpointer key;
   g_autoptr(GError) local_error = NULL;
-
-  if (opt_arch != NULL)
-    arches = opt_arches;
 
   if (opt_name && !flatpak_is_valid_name (opt_name, &local_error))
     {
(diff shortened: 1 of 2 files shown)


common/flatpak-dir.c
```diff
@@ -7935,6 +7935,61 @@ typedef enum {
...
+  if (opt_name != NULL && g_strcmp0 (opt_name, parts[1]) != 0)
+    return;
+
+  if (!g_strv_contains (arches, parts[2]))
+    return;
+
+  if (opt_branch != NULL && g_strcmp0 (opt_branch, parts[3]) != 0)
+    return;
+
+  if (opt_collection_id != NULL && g_strcmp0 (opt_collection_id, coll_ref->collection_id))
```
cosimoc: Nitpick: add `!= 0` at the end like in the other blocks

cosimoc: Nitpick: add `!= 0` at the end like in the other blocks

common/flatpak-dir.c
```diff
@@ -7973,50 +8023,64 @@ find_matching_refs (GHashTable *refs,
...
+      g_autoptr(GVariant) child = NULL;
+      g_autoptr(GVariant) cur_v = NULL;
+      g_autoptr(FlatpakCollectionRef) coll_ref = NULL;
+      const char *ref;
 
-      if (opt_collection_id != NULL && strcmp (opt_collection_id, coll_ref->collection_id))
-        continue;
+      child = g_variant_get_child_value (cache, i);
+      cur_v = g_variant_get_child_value (child, 0);
+      ref = g_variant_get_data (cur_v);
```
cosimoc: I commented this before, but I believe it would be best to use the typed method to get this data; e.g. `g_variant_get_string()` or `g_variant_get_bytestring()` (not sure what the type of this variant is).

common/flatpak-dir.c
```diff
@@ -8223,27 +8309,85 @@ flatpak_dir_find_remote_ref (FlatpakDir   *self,
...
+                *out_kind = FLATPAK_KINDS_APP;
+              else
+                *out_kind = FLATPAK_KINDS_RUNTIME;
+            }
         }
       else
         {
-          g_propagate_error (error, g_steal_pointer (&my_error));
+          g_autoptr(GString) err = g_string_new ("");
+          g_string_printf (err, _("Multiple branches available for %s, you must specify one of: "), name);
```
cosimoc: The code at https://github.com/endlessm/flatpak/blob/47536627212bd3f60647ce49f943ec2e7fcba890/common/flatpak-dir.c#L8124 looks at `opt_default_branch` to determine if one of these is the one we're looking for. Perhaps we should do the same here, and factor that code out into a helper function.

common/flatpak-dir.c
```diff
@@ -8128,21 +8203,43 @@ flatpak_dir_find_remote_refs (FlatpakDir   *self,
...
+    }
+  else
+    {
+      g_autoptr(GVariant) xa_cache = NULL;
+
+      /* Use xa.cache because we can't fetch the summary when offline */
+      if (!flatpak_dir_fetch_refs_cache (self, remote, &xa_cache, cancellable, error))
+        return NULL;
+
+      matched_refs = find_matching_refs_from_cache (xa_cache,
```
pwithnall: Is there a way you could merge `find_matching_refs()` and `find_matching_refs_from_cache()`? These two branches (for `collection == NULL`) are really quite similar, and I suspect they’re going to have to be replicated wherever `find_matching_refs()` is called.

Alternatively, you could factor out this whole `if (collection_id == NULL) … else …` block into a function which returns `matched_refs`.

common/flatpak-dir.c
```diff
@@ -10447,17 +10612,24 @@ flatpak_dir_find_remote_related (FlatpakDir *self,
 
               extension_ref = g_build_filename ("runtime", extension, parts[2], branch, NULL);
 
-              if (flatpak_summary_lookup_ref (summary, extension_collection_id, extension_ref, &checksum, NULL))
+              if ((summary == NULL && flatpak_cache_lookup_ref (xa_cache, extension_ref)) ||
+                  (summary != NULL && flatpak_summary_lookup_ref (summary, extension_collection_id, extension_ref, &checksum, NULL)))
```
pwithnall: Would it be possible to merge `flatpak_cache_lookup_ref()` and `flatpak_summary_lookup_ref()`? Otherwise you’re going to have to conditionally call both of them from every call site. This should be possible if you can pass the same data structure to both functions (regardless of whether it came from a `summary` file or `ostree-metadata`).

common/flatpak-dir.c
```diff
@@ -10447,17 +10612,24 @@ flatpak_dir_find_remote_related (FlatpakDir *self,
...
                   add_related (self, related, extension, extension_collection_id, extension_ref, checksum, no_autodownload, download_if, autodelete, locale_subset);
                 }
               else if (subdirectories)
                 {
-                  g_auto(GStrv) refs = flatpak_summary_match_subrefs (summary, extension_collection_id, extension_ref);
+                  g_auto(GStrv) refs;
+                  if (summary != NULL)
+                    refs = flatpak_summary_match_subrefs (summary, extension_collection_id, extension_ref);
+                  else
+                    refs = flatpak_cache_match_subrefs (xa_cache, extension_ref);
```
pwithnall: Similarly, it would be good to merge `flatpak_summary_match_subrefs()` and `flatpak_cache_match_subrefs()`.

common/flatpak-utils.c
```diff
@@ -2711,63 +2711,87 @@ summary_find_refs_list (GVariant   *summary,
   return g_steal_pointer (&refs);
 }
 
-/* This matches all refs from @collection_id that have ref, followed by '.'  as prefix */
-char **
-flatpak_summary_match_subrefs (GVariant   *summary,
-                               const char *collection_id,
-                               const char *ref)
+/* Note: This function relies on the similar structure in both xa.cache and summary refs lists */
```
pwithnall: Similar? Or identical? If they’re not identical then I think this comment should give the `GVariantType` for both and make the differences more obvious, otherwise it’s not much use to anyone who has to modify this function in future.

pwithnall: In any case, you should give the expected `GVariantType` of `haystack`.

common/flatpak-dir.c
```diff
@@ -10599,17 +10612,24 @@ flatpak_dir_find_remote_related (FlatpakDir *self,
 
               extension_ref = g_build_filename ("runtime", extension, parts[2], branch, NULL);
 
-              if (flatpak_summary_lookup_ref (summary, extension_collection_id, extension_ref, &checksum, NULL))
+              if ((summary == NULL && flatpak_cache_lookup_ref (xa_cache, extension_ref)) ||
```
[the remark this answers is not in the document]
mwleeds: As I said in the commit message, it doesn't matter for the callers that I checked (it's public API so there could be others). I left the parameter so the behavior would stay the same for anyone not using collection IDs.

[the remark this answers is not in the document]
mwleeds: (This function isn't public API but callers of it are)

common/flatpak-dir.c
mwleeds: Not really possible to merge them, because they're iterating over different data structures. I suppose you could have the caller pass one or the other (the xa.cache or the summary), but that seems ugly. I don't think factoring out the whole block is necessary either because this is the only place it's used.

common/flatpak-utils.c
mwleeds: similar

mwleeds: I don't see how I can give a type when it could be two different types, and someone modifying it should find those types from the source anyway.

pwithnall:
```
/* Note: This function relies on the similar structure of both `xa.cache` (type: `a{s(tts)}`) and summary refs lists (type: `a(s(taya{sv}))`). */
```

common/flatpak-dir.c
pwithnall: Introspection would be terrible. Relying on the data structure being similar enough for `flatpak_variant_bsearch_str()` is pretty yucky too, but will do for a first go at the branch.

common/flatpak-dir.c
pwithnall: I am worried that someone will write some code in flatpak in future which just calls `flatpak_summary_match_subrefs()`, and forgets to handle the `flatpak_cache_match_subrefs()` case too, leading to a regression in P2P support. Similarly with the other functions here. When this is submitted upstream, I strongly suggest you find a way to factor them together (or factor this out into a higher level function `flatpak_repo_metadata_match_subrefs()` which grabs the `summary` or `ostree-metadata` and matches as appropriate), or there will be regressions in future.

mwleeds: That sounds like a valid concern; thanks for explaining. I'll think about what can be done. Cosimo said this won't be ready in time for the demo anyway, so maybe we should hold off on merging it until we get something upstream.

PR #81, opened 2018-04-19
Update xenial to 0.10.4

dbnicholson:
This is our master branch rebased onto 0.10.4 with one more patch debian was carrying to not require fuse if the document portal isn't being built. Our changes relative to 0.10.4 can be seen at https://github.com/endlessm/flatpak/compare/0.10.4...T22022-flatpak-0.10.4. I didn't make any effort to see if any of those patches could be left out on the infrastructure.

Please don't merge this as I'll push all the branches and tags at one time.

https://phabricator.endlessm.com/T22022

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4786,10 +4786,13 @@ get_permissible_prefixes (FlatpakContext                  *context,
 
       *out_match_prefixes_func = flatpak_name_matches_one_wildcard_prefix;
     }
+  else
+    {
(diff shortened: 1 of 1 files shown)


smspillaz:
We will also need to pick the upstream changes to service file exports and tagged runtimes. Or should I do that myself once this branch has been merged?

smspillaz:
FWIW I have a branch here: https://github.com/endlessm/flatpak/tree/T22022-flatpak-0.10.4-T22077-T22106 which includes everything that is needed from my side. You should just be able to cherry-pick the first 12 commits directly

PR #83, opened 2018-04-19
Upstream master (DON'T MERGE)

mwleeds:
This PR only exists so I can see what our CI thinks of upstream master.


mwleeds:
It appears that our CI doesn't enable P2P. I'll see if I can make a commit on this branch to change that.

PR #84, opened 2018-04-20
Enable P2P by default

mwleeds:
This commits makes `--enable-p2p` default to yes rather than no, so that
the CI uses it when running tests (since we already use it for the
version of flatpak that goes into Endless). This also has the effect of
making it easier for people who build Flatpak to have a version closer
to what goes into the OS.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -263,8 +263,8 @@ AC_SUBST([SUDO_BIN])
 # included.
 AC_ARG_ENABLE([p2p],
   [AS_HELP_STRING([--enable-p2p],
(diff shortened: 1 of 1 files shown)


alexlarsson:
I don't want to ack things in the endless repo. But I don't want to merge this to the public repo until the p2p API is on by default in ostree.

pwithnall:
I swear I read this as being in flatpak/flatpak rather than endlessm/flatpak. More :coffee: needed.

In that case, @mwleeds, it’s up to you whether you merge now or add another commit to change the `<!-- FIXME -->`s.

PR #87, opened 2018-04-23
Update to 0.11.3-3 Debian packaging (DON'T MERGE)

mwleeds:
This imports some changes from flatpak_0.11.3-3.debian.tar.xz and adds a
changelog entry for the new downstream release (just the current state
of upstream master plus 8 patches).

https://phabricator.endlessm.com/T21153


mwleeds:
I've never dealt with Debian packaging before, so review with skepticism.

cosimoc:
This looks fine to me. It would probably be OK to keep the python3 test stuff from Debian, but I'm fine waiting for that to go upstream too.

debian/changelog
```diff
@@ -1,3 +1,96 @@
+flatpak (0.11.3-3endless0) unstable; urgency=medium
```
cosimoc: Should be `endless1`

debian/changelog
```diff
@@ -1,3 +1,96 @@
...
+  * New upstream releases
+  * Rebase the eight Endless patches
```
cosimoc: The debian packaging does not know about these patches; I'd rather drop this comment.

debian/changelog
```diff
@@ -1,3 +1,96 @@
+flatpak (0.11.3-3endless0) unstable; urgency=medium
+  * New upstream releases
```
cosimoc: An alternative approach here would be to:
- reset the whole changelog to debian 0.11.3
- add on top our own changelog entry with a description of the changes that we still carry on top of debian packaging

I find that easier when rebasing and having to consider whether changes are still required going forward, but Iguess it's ultimately up to your preference.

debian/changelog
mwleeds: Further down there's a `0.10.1-1endless0` release. Was that wrong?

cosimoc: Yeah... there's nothing *wrong* in the sense that stuff would break, but I believe that we usually count from 1 for that version.

debian/changelog
mwleeds: Sure

debian/changelog
mwleeds: I think having the whole history makes sense.

pwithnall:
Jenkins is failing because it’s building the `master` branch together with this Debian branch. From [the log](https://ci.endlessm-sf.com/job/obs-flatpak-pr/45/console):
```
+ jenkins-obs-pr
Building endlessm/flatpak origin/pr/87/merge: merge of mwleeds/T21153-0.11.3-packaging to debian-master
No corresponding source branch found; using 'master'
```

Jenkins should be able to associate the two branches and build them together if you name them `T21153-0.11.3` and `T21153-0.11.3-debian` (specifically, it looks for branches with/without the `-debian` suffix).

As a more general review point though, I think you should consider changing our flatpak `debian-master` branch to be rebased on the [upstream Debian packaging git branch](https://salsa.debian.org/debian/flatpak), since Debian now keeps its packaging in git. If we base our `debian-master` on their packaging branch, rebases should actually be a matter of `git rebase`. See what Dan did for the Xenial packaging (#82) or the OSTree packaging (https://github.com/endlessm/ostree/pull/115).

However, if you’re pushed for time, what you have here looks OK to me. [Here’s the diff between salsa/debian/0.11.3-3 and this branch](https://phabricator.endlessm.com/P535), which looks convincing. As does [the diff between salsa/debian/0.10.1-1 and the current debian-master branch](https://phabricator.endlessm.com/P536).

mwleeds:
I pushed this.

PR #88, opened 2018-04-24
utils: Don't pass a null out-parameter to flatpak_variant_bsearch_str

smspillaz:


mwleeds:
That branch won't be merged anyway because I'll have to rewrite it soon, but sure.

PR #92, opened 2018-05-01
flatpak-dir: make optional summary really optional

cosimoc:
We should not require a collection ID to be set for this to actually be
optional.
Doing so will make us require a network connection for flaptak installs
and updates even in case we already have the requested ref available,
for instance because it has been previously pulled by a different process,
as it's the case during an Endless OS update that also involves
updating or installing applications.

https://phabricator.endlessm.com/T22362

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -8284,7 +8284,7 @@ _flatpak_dir_get_remote_state (FlatpakDir   *self,
         }
       else
         {
-          if (optional && state->collection_id != NULL)
+          if (optional)
             {
               g_debug ("Failed to download optional summary");
             }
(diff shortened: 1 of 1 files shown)


pwithnall:
Nitpick: s/flaptak/flatpak/ in the commit message.

pwithnall:
I think this is definitely @mwleeds’ domain. I have no idea about the intended behaviour of optional downloads, and am unable to find the code review discussion where they were discussed. That code was introduced by https://github.com/flatpak/flatpak/pull/1575.

mwleeds:
It would be good to update the comment above `flatpak_dir_get_remote_state_optional()` to take out `in the p2p case` and add a sentence about how a no-pull install is another use case for it.

mwleeds:
And it's more in line with convention to start the commit message `dir:`

mwleeds:
It also might make sense to be specific about Endless' use case in the commit message, and say it's calling `flatpak_installation_install` with `FLATPAK_UPDATE_FLAGS_NO_PULL` set.

mwleeds:
Other than that and pending your check to see if this fixes the problem, LGTM

mwleeds:
Never mind, we can keep this one on topic and use https://github.com/flatpak/flatpak/pull/1637 for the disabled remote issue.

cosimoc:
Updated the commit message and merged this. @mwleeds is going to take care of bringing this upstream.

PR #95, opened 2018-05-22
Allow requesting "no prune" updates/uninstalls via the public API

mariospr:
This will be consumed by eos-updater for now, but could be interesting for upstream too.

https://phabricator.endlessm.com/T22621

### DIFF
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -1897,8 +1897,8 @@ flatpak_installation_update_full (FlatpakInstallation    *self,
   if (result == NULL)
     goto out;
 
-  /* We don't get prunable objects if not pulling */
-  if (!(flags & FLATPAK_UPDATE_FLAGS_NO_PULL))
+  /* We don't get prunable objects if not pulling or if NO_PRUNE is passed */
+  if (!(flags & FLATPAK_UPDATE_FLAGS_NO_PULL) && !(flags & FLATPAK_UPDATE_FLAGS_NO_PRUNE))
     flatpak_dir_prune (dir_clone, cancellable, NULL);
 
 out:
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -1897,7 +1897,8 @@ flatpak_installation_update_full (FlatpakInstallation    *self,
   if (result == NULL)
     goto out;
 
-  if ((flags & FLATPAK_UPDATE_FLAGS_NO_DEPLOY) != 0)
+  /* We don't get prunable objects if not pulling */
+  if (!(flags & FLATPAK_UPDATE_FLAGS_NO_PULL))
     flatpak_dir_prune (dir_clone, cancellable, NULL);
 
 out:
--- a/lib/flatpak-installation.c
+++ b/lib/flatpak-installation.c
@@ -1978,6 +1978,43 @@ flatpak_installation_uninstall (FlatpakInstallation    *self,
                                 GCancellable           *cancellable,
                                 GError                **error)
 {
+  return flatpak_installation_uninstall_full (self, FLATPAK_UNINSTALL_FLAGS_NONE,
+                                              kind, name, arch, branch,
+                                              progress, progress_data,
+                                              cancellable, error);
+}
+
+/**
+ * flatpak_installation_uninstall_full:
+ * @self: a #FlatpakInstallation
+ * @flags: set of #FlatpakUninstallFlags flags
+ * @kind: what this ref contains (an #FlatpakRefKind)
+ * @name: name of the app or runtime to uninstall
+ * @arch: architecture of the app or runtime to uninstall
+ * @branch: name of the branch of the app or runtime to uninstall
+ * @progress: (scope call) (nullable): the callback
+ * @progress_data: (closure progress) (nullable): user data passed to @progress
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Uninstall an application or runtime.
+ *
+ * Returns: %TRUE on success
+ *
+ * Since: 0.11.8
+ */
+gboolean
+flatpak_installation_uninstall_full (FlatpakInstallation    *self,
+                                     FlatpakUninstallFlags   flags,
+                                     FlatpakRefKind          kind,
+                                     const char             *name,
+                                     const char             *arch,
+                                     const char             *branch,
+                                     FlatpakProgressCallback progress,
+                                     gpointer                progress_data,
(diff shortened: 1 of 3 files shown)


lib/flatpak-installation.c
```diff
@@ -1897,7 +1897,8 @@ flatpak_installation_update_full (FlatpakInstallation    *self,
   if (result == NULL)
     goto out;
 
-  if ((flags & FLATPAK_UPDATE_FLAGS_NO_DEPLOY) != 0)
```
cosimoc: I would not be surprised if this was actually a bug in the upstream logic. Can you double check with @alexlarsson?
If we're doing a `NO_DEPLOY` update, it means we're only downloading new objects for an update, not undeploying the old version, and not deploying the new one.
So there should be nothing to prune in that case, as opposed to when we actually do a regular update or a `NO_PULL` one, as that will undeploy the old version which may result in objects being available for pruning...

lib/flatpak-installation.h
```diff
@@ -76,6 +76,17 @@ typedef enum {
   FLATPAK_INSTALL_FLAGS_NO_PULL          = (1 << 3),
 } FlatpakInstallFlags;
 
+/**
+ * FlatpakUninstallFlags:
+ * @FLATPAK_UNINSTALL_FLAGS_NONE: Default
```
cosimoc: Missing doc string for the NO_PRUNE flag

lib/flatpak-installation.c
```diff
@@ -1897,7 +1897,8 @@ flatpak_installation_update_full (FlatpakInstallation    *self,
...
+  if ((flags & FLATPAK_UPDATE_FLAGS_NO_DEPLOY) != 0 &&
+      (flags & FLATPAK_UPDATE_FLAGS_NO_PRUNE) == 0)
```
pwithnall: Nitpick: I’d say `!(flags & FLATPAK_UPDATE_FLAGS_NO_PRUNE)` is a bit more obvious to read, since the sense of the condition is not left to the end of the statement. However, it depends on the prevailing flatpak coding style.

lib/flatpak-installation.c
```diff
@@ -1976,6 +1977,41 @@ flatpak_installation_uninstall (FlatpakInstallation    *self,
...
+ * @arch: architecture of the app or runtime to uninstall
+ * @branch: name of the branch of the app or runtime to uninstall
+ * @progress: (scope call) (nullable): the callback
+ * @progress_data: (closure progress) (nullable): user data passed to @progress
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Uninstall an application or runtime.
+ *
+ * Returns: %TRUE on success
```
pwithnall: Nitpick: Needs a `Since: xx` line giving the version number of the first release which would contain this new API.

lib/flatpak-installation.c
```diff
@@ -1976,6 +1977,41 @@ flatpak_installation_uninstall (FlatpakInstallation    *self,
...
+  return flatpak_installation_uninstall_full (self, FLATPAK_UNINSTALL_FLAGS_NONE,
+                                              kind, name, arch, branch,
+                                              progress, progress_data,
+                                              cancellable, error);
+}
+
+/**
+ * flatpak_installation_uninstall_full:
+ * @self: a #FlatpakInstallation
+ * @flags: set of #FlatpakUninstallFlags flag
```
pwithnall: Nitpick: s/flag/flags/

lib/flatpak-installation.c
```diff
@@ -1976,6 +1977,41 @@ flatpak_installation_uninstall (FlatpakInstallation    *self,
...
+ * @kind: what this ref contains (an #FlatpakRefKind)
```
pwithnall: No need to mention this is a `FlatpakRefKind`; gtk-doc shows the type already, separately.

lib/flatpak-installation.c
```diff
@@ -1998,7 +2034,8 @@ flatpak_installation_uninstall (FlatpakInstallation    *self,
                               cancellable, error))
     return FALSE;
 
-  flatpak_dir_prune (dir_clone, cancellable, NULL);
+  if ((flags & FLATPAK_UNINSTALL_FLAGS_NO_PRUNE) == 0)
```
pwithnall: Nitpick: Same comment about using `!` rather than `== 0` here.

lib/flatpak-installation.h
```diff
@@ -75,6 +76,17 @@ typedef enum {
...
+ *
+ * Flags to alter the behavior of flatpak_installation_uninstall_full().
```
pwithnall: Nitpick: Needs a `Since: xx` line too.

lib/flatpak-installation.c
pwithnall: Yeah, agreed. We talked about this in London today, and we also concluded that the prune operations which flatpak does should be conditional on `!NO_PULL`, rather than anything to do with `NO_DEPLOY`. flatpak’s pruning is based on reachability from refs, and deploying/not deploying an app doesn’t affect its refs. Pulling/Not pulling an app does.

pwithnall: This likely needs fixing in other places in flatpak too.

pwithnall:
This needs submitting upstream too, and ideally we would have some upstream feedback from Alex about whether the API is acceptable before we start depending on it in eos-updater. I don’t want to diverge from upstream if at all possible.

lib/flatpak-installation.c
mariospr: I double checked with him and I believe he'd be fine with such a change. I'm working on it now

lib/flatpak-installation.h
mariospr: Ok

lib/flatpak-installation.c
mariospr: Sure

lib/flatpak-installation.c
mariospr: This comes from the original `flatpak_installation_uninstall()` method, not really my wording.

lib/flatpak-installation.c
mariospr: Same reply: let's see what ends up being merged upstream, and go with that

lib/flatpak-installation.h
mariospr: Sure

mariospr:
This has been merged in https://github.com/flatpak/flatpak/pull/1704 and https://github.com/flatpak/flatpak/pull/1705 upstream, so I've just updated this PR to match the work there.

PR #97, opened 2018-05-29
common: Pop the thread default main context in find_latest_rev()

pwithnall:
The pop was missing, meaning that calling flatpak_dir_find_latest_rev()
corrupted the thread default main context stack of the caller.


pwithnall:
This was causing eos-updater to reliably crash for me in testing. I need to do some analysis to work out the extent of the impact (if any).

pwithnall:
Also merged upstream: https://github.com/flatpak/flatpak/pull/1731

PR #98, opened 2018-06-21
dir: Never fsync child repos

ramcq:
There is no need to force a fsync after pulling into the child repo,
because we will anyway copy/verify it into the system repo. It is
never used for stable storage.

This makes system installation faster.

Closes: #1808
Approved by: alexlarsson

(cherry picked from commit 9901ce875fc88db49ec3f6e72ffa850ab3d22ac9)
Signed-off-by: Robert McQueen <rob@endlessm.com>

https://phabricator.endlessm.com/T23081


ramcq:
https://phabricator.endlessm.com/T23081

PR #100, opened 2018-06-28
Uajain/min free space size

uajain:
https://phabricator.endlessm.com/T13698

This PR should be reviewed in conjunction with https://github.com/endlessm/ostree/pull/126

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1958,8 +1958,8 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
             }
         }
 
-      /* Reset min-free-space-percent to 0, this keeps being a problem for a lot
-       * of people.
+      /* Earlier flatpak used to reset min-free-space-percent to 0 everytime, but now we
+       * favor min-free-space-size  instead of it (See below).
        *
        * Similarly (Endless only), ensure core.add-remotes-config-dir defaults
        * to false, otherwise we can end up writing partially valid config files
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -1970,18 +1970,29 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
...
-          if (orig_min_free_space_percent == NULL ||
+          if (orig_min_free_space_size == NULL ||
               orig_add_remotes_config_dir == NULL)
             new_config = ostree_repo_copy_config (repo);
 
-          if (orig_min_free_space_percent == NULL)
-            g_key_file_set_string (new_config, "core", "min-free-space-percent", "0");
+          /* Scrap previously written min-free-space-percent=0 and replace it with min-free-space-size */
+          if (orig_min_free_space_percent != NULL &&
+              g_ascii_strtoull (orig_min_free_space_percent, NULL, 0) == 0)
```
pwithnall: `g_ascii_strtoull()` returns `0` on error, so this condition will also hold true if there’s an error parsing the setting.

You might want to use `g_ascii_string_to_unsigned()` instead, if flatpak depends on a recent enough version of GLib. It simplifies the error handling.

common/flatpak-dir.c
```diff
@@ -1970,18 +1970,29 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
...
+            {
+              g_key_file_remove_key (new_config, "core", "min-free-space-percent", NULL);
+              g_key_file_set_string (new_config, "core", "min-free-space-size", "500MB");
```
pwithnall: Nitpick: Might want to factor out the `500MB` default from here and below into a common `const` variable above.

common/flatpak-dir.c
```diff
@@ -1970,18 +1970,29 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
...
           g_autofree char *orig_add_remotes_config_dir = NULL;
 
           orig_config = ostree_repo_get_config (repo);
           orig_min_free_space_percent = g_key_file_get_value (orig_config, "core", "min-free-space-percent", NULL);
+          orig_min_free_space_size = g_key_file_get_value (orig_config, "core", "min-free-space-size", NULL);
           orig_add_remotes_config_dir = g_key_file_get_value (orig_config, "core", "add-remotes-config-dir", NULL);
 
-          if (orig_min_free_space_percent == NULL ||
+          if (orig_min_free_space_size == NULL ||
               orig_add_remotes_config_dir == NULL)
```
pwithnall: This condition needs to include `|| orig_min_free_space_percent != NULL`, otherwise `new_config` will still be set to `NULL` if the `if (orig_min_free_space_percent != NULL && g_ascii_strtoull (…) == 0)` branch is taken below.

common/flatpak-dir.c
```diff
@@ -1970,18 +1970,42 @@ _flatpak_dir_ensure_repo (FlatpakDir   *self,
...
+          if (orig_min_free_space_size == NULL ||
+              orig_min_free_space_percent != NULL ||
               orig_add_remotes_config_dir == NULL)
             new_config = ostree_repo_copy_config (repo);
 
-          if (orig_min_free_space_percent == NULL)
-            g_key_file_set_string (new_config, "core", "min-free-space-percent", "0");
+          /* Scrap previously written min-free-space-percent=0 and replace it with min-free-space-size */
+          if (orig_min_free_space_percent != NULL &&
+              g_ascii_string_to_unsigned (orig_min_free_space_percent, 10,
```
pwithnall: If this is going to be upstreamed, note that you’ll need to bump the `GLIB_REQS` in `configure.ac` to `2.54.0`, since that’s when those APIs were added to GLib.

That’s not necessary on Endless, since we’ve backported the APIs to our current version of GLib.

PR #102, opened 2018-07-18
Rebase to upstream 0.99.3 (DON'T MERGE)

mwleeds:
This is the current state of upstream master, plus the 10 downstream commits we're carrying. Everything else we have is already upstream. The packaging branch for this and the ostree rebase still need to be done before this can be pushed.

https://phabricator.endlessm.com/T23138

### DIFF
--- a/profile/flatpak.sh.in
+++ b/profile/flatpak.sh.in
@@ -1,7 +1,13 @@
-# @sysconfdir@/profile.d/flatpak.sh - set XDG_DATA_DIRS
+# @sysconfdir@/profile.d/flatpak.sh - set XDG_DATA_DIRS and PATH
 
 if [ "${XDG_DATA_DIRS#*flatpak}" = "${XDG_DATA_DIRS}" ]; then
     XDG_DATA_DIRS="${XDG_DATA_HOME:-"$HOME/.local/share"}/flatpak/exports/share:@localstatedir@/lib/flatpak/exports/share:@externalinstalldir@/exports/share/:${XDG_DATA_DIRS:-/usr/local/share:/usr/share}"
 fi
 
 export XDG_DATA_DIRS
+
+if [ "${PATH#*flatpak}" = "${PATH}" ]; then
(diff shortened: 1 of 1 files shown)


mwleeds:
There were only a few conflicts and they're documented in the commit messages.

Makefile.am
```diff
@@ -41,6 +41,7 @@ ACLOCAL_AMFLAGS = -I m4 -I libglnx ${ACLOCAL_FLAGS}
 AM_CPPFLAGS =							\
 	-DFLATPAK_BINDIR=\"$(FLATPAK_BINDIR)\"			\
 	-DFLATPAK_SYSTEMDIR=\"$(SYSTEM_INSTALL_DIR)\"		\
+	-DFLATPAK_EXTERNALDIR=\"$(EXTERNAL_INSTALL_DIR)\"	\
```
cosimoc: Typo in the commit message: "defining and external" -> "defining an external"

Makefile.am
```diff
@@ -25,6 +25,7 @@ SUBDIRS += po
 %.service: %.service.in config.log
```
cosimoc: The latter paragraph in the commit message does not seem to apply to the code at hand anymore.

uajain:
@mwleeds yes, those patches are very much intended to go upstream. The only concern here is that to make it in upstream flatpak, a new release of ostree is required.

So what I have now is:
Release PR for OStree: https://github.com/ostreedev/ostree/pull/1683
Upstream flatpak patches PR : https://github.com/flatpak/flatpak/pull/1912

Makefile.am
mwleeds: Fixed

Makefile.am
cosimoc: Yeah, the paragraph I refer to was
```
Also, make sure to run with XDG_DATA_DIRS set. Otherwise we can see in
the log that update-mime-database will complain about it not being set
when installing an application using the system helper.
```

As far as I can see, the helper already runs with `XDG_DATA_DIRS` set; this patch only adds a new path to it.

mwleeds: I edited the commit message. It looks like it's been inaccurate since at least the 3.0.4 release so maybe it was always that way.

mwleeds:
Here are my notes from the rebase in case they're helpful:
[flatpak_0.99.3_rebase.csv.txt](https://github.com/endlessm/flatpak/files/2211725/flatpak_0.99.3_rebase.csv.txt)

PR #103, opened 2018-07-20
Update packaging for 0.99.3 rebase (DON'T MERGE)

mwleeds:
I started with the upstream Debian tag `debian/0.99.3-1` from https://salsa.debian.org/debian/flatpak and did a recursive diff to look at the differences between that and the `debian` directory we have on `debian-master`. It turned out that there's only one patch we need to keep (see the changelog for details).

This is supposed to be a PR against the `debian-master` branch but GitHub won't let me do that since they have different histories.

I also noticed that the bubblewrap requirement is now 0.2.1 and we have 0.2.0 in Endless so that upgrade will block this in addition to the ostree upgrade.

https://phabricator.endlessm.com/T23138


mwleeds:
Merged this

PR #105, opened 2018-08-08
0.99.3 upstream cherry-picks

mwleeds:
These are commits from upstream that seem important, as well as a typo fix commit that prevents merge conflicts.


ramcq:
Alex just released 0.99.4. As we're at the very start of the cycle, maybe we should just fast-forward to 0.99.4? Nothing looks that scary.

mwleeds:
Sure, we can do that. The delta is pretty trivial.

PR #107, opened 2018-08-16
RFC: Support migration from old app-ids to new app-id

uajain:
flatpak run checks if there is any previous-ids for the app and
sees if there is a corresponding app-dir in ~/.var/app/ .
If so, it renames that app-dir to the new app-id and provides
a symlink from old app-id dir to new app-id dir. The symlink is
needed because even though the XDG_DATA_DIR and other environment
variables are set to the new ID by flatpak run, the app itself
could've stored (eg in GSettings, or other files) paths which
contained the old ID.

https://phabricator.endlessm.com/T23532

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -1935,6 +1935,25 @@ flatpak_deploy_data_get_eol_rebase (GVariant *deploy_data)
   return eol;
 }
 
+const char **
+flatpak_deploy_data_get_previous_ids (GVariant *deploy_data, gsize *length)
+{
+  g_autoptr(GVariant) metadata = g_variant_get_child_value (deploy_data, 4);
+  g_autoptr(GVariant) previous_ids_v = NULL;
+  const char **previous_ids = NULL;
+
+  previous_ids_v = g_variant_lookup_value (metadata, "previous-ids", NULL);
+  if (previous_ids_v)
+    previous_ids = g_variant_get_strv (previous_ids_v, length);
+  else
+    {
+      if (length != NULL)
+        *length = 0;
+    }
+
+  return previous_ids;
+}
+
 const char *
 flatpak_deploy_data_get_runtime (GVariant *deploy_data)
 {
@@ -6990,6 +7009,13 @@ flatpak_dir_deploy (FlatpakDir          *self,
   if (application_runtime)
     g_variant_builder_add (&metadata_builder, "{s@v}", "runtime",
                            g_variant_new_variant (g_variant_new_string (application_runtime)));
+  if (old_deploy_data)
+    {
+      g_autofree const char **previous_ids = flatpak_deploy_data_get_previous_ids (old_deploy_data, NULL);
+      if (previous_ids)
+        g_variant_builder_add (&metadata_builder, "{s@v}", "previous-ids",
+                               g_variant_new_variant (g_variant_new_strv ((const char * const *) previous_ids, -1)));
+    }
 
   deploy_data = flatpak_dir_new_deploy_data (origin,
                                              checksum,
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2963,7 +2963,45 @@ flatpak_run_app (const char     *app_ref,
 
   if (app_deploy != NULL)
     {
+      g_autofree const char **previous_ids = NULL;
+      gsize len = 0;
+
       app_files = flatpak_deploy_get_files (app_deploy);
+
+      g_autoptr (GFile) current_data_dir = flatpak_get_data_dir (app_ref_parts[1]);
+      previous_ids = flatpak_deploy_data_get_previous_ids (app_deploy_data, &len);
+      while (len > 0)
+        {
+          g_autoptr (GFile) previous_data_dir = NULL;
+          g_autoptr (GError) local_error = NULL;
+          g_autofree char *previous_data_path = NULL;
+          g_autofree char *current_data_path = NULL;
+
+          if (g_file_query_exists (current_data_dir, cancellable))
+            break;
+
+          previous_data_dir = flatpak_get_data_dir (previous_ids[--len]);
+          if (!g_file_query_exists (previous_data_dir, cancellable))
+            continue;
+
+          if (!flatpak_file_rename (previous_data_dir,
+                                    current_data_dir,
+                                    cancellable, &local_error))
+             {
+               g_propagate_error (error, g_steal_pointer (&local_error));
+               return FALSE;
+             }
+
+           current_data_path = g_file_get_path (current_data_dir);
+           previous_data_path = g_file_get_path (previous_data_dir);
+           if (symlink (current_data_path, previous_data_path) == 0)
+             break;
+           else
+             {
+               glnx_set_error_from_errno (error);
+               return FALSE;
+             }
+        }
 
       real_app_id_dir = flatpak_ensure_data_dir (app_ref_parts[1], cancellable, error);
       if (real_app_id_dir == NULL)
--- a/common/flatpak-dir-private.h
+++ b/common/flatpak-dir-private.h
@@ -233,6 +233,9 @@ const char *        flatpak_deploy_data_get_alt_id (GVariant *deploy_data);
 const char *        flatpak_deploy_data_get_eol (GVariant *deploy_data);
 const char *        flatpak_deploy_data_get_eol_rebase (GVariant *deploy_data);
 const char *        flatpak_deploy_data_get_runtime (GVariant *deploy_data);
+const char **       flatpak_deploy_data_get_previous_ids (GVariant *deploy_data,
+                                                          gsize    *length);
+
 
 GFile *        flatpak_deploy_get_dir (FlatpakDeploy *deploy);
 GVariant *     flatpak_load_deploy_data (GFile        *deploy_dir,


common/flatpak-run.c
```diff
@@ -2961,8 +2961,40 @@ flatpak_run_app (const char     *app_ref,
...
+      if (previous_ids)
+        {
+          for (i = 0; previous_ids[i] != NULL; i++)
+            {
+              g_autoptr (GFile) previous_data_dir = NULL;
+
+              previous_data_dir = flatpak_get_data_dir (previous_ids[i]);
+              if (g_file_query_exists (previous_data_dir, NULL))
+                {
+                  g_autoptr (GFile) current_data_dir = flatpak_get_data_dir (app_ref_parts[1]);
```
ramcq: get this dir once at the start (outside the loop)

common/flatpak-run.c
```diff
@@ -2961,8 +2961,40 @@ flatpak_run_app (const char     *app_ref,
...
 
+      previous_ids = flatpak_deploy_data_get_previous_ids (app_deploy_data);
+      if (previous_ids)
+        {
+          for (i = 0; previous_ids[i] != NULL; i++)
+            {
+              g_autoptr (GFile) previous_data_dir = NULL;
+
+              previous_data_dir = flatpak_get_data_dir (previous_ids[i]);
+              if (g_file_query_exists (previous_data_dir, NULL))
```
ramcq: you can do this check backwards, like if !exists; continue

common/flatpak-run.c
```diff
@@ -2961,8 +2961,40 @@ flatpak_run_app (const char     *app_ref,
 
   if (app_deploy != NULL)
     {
+      g_autofree const char **previous_ids = NULL;
+
       app_files = flatpak_deploy_get_files (app_deploy);
 
+      previous_ids = flatpak_deploy_data_get_previous_ids (app_deploy_data);
```
ramcq: we should reverse the list so that we use the most recent ID first, if it exists

common/flatpak-run.c
```diff
@@ -2961,8 +2961,40 @@ flatpak_run_app (const char     *app_ref,
...
+      if (previous_ids)
+        {
+          for (i = 0; previous_ids[i] != NULL; i++)
+            {
+              g_autoptr (GFile) previous_data_dir = NULL;
+
```
ramcq: we should check at the start of the loop (each time around, because a previous iteration can create it) whether the current dir exists - if it does, we should break; and we have no more work to do

common/flatpak-run.c
```diff
@@ -2961,8 +2961,40 @@ flatpak_run_app (const char     *app_ref,
...
+              previous_data_dir = flatpak_get_data_dir (previous_ids[i]);
+              if (g_file_query_exists (previous_data_dir, NULL))
+                {
+                  g_autoptr (GFile) current_data_dir = flatpak_get_data_dir (app_ref_parts[1]);
+                  g_autoptr (GError) local_error = NULL;
+
+                  if (flatpak_file_rename (previous_data_dir,
```
ramcq: you can ! this as well, so just do the error checking and return early if there is a problem

common/flatpak-run.c
```diff
@@ -2961,8 +2961,40 @@ flatpak_run_app (const char     *app_ref,
...
+                                           current_data_dir,
+                                           cancellable, &local_error))
+                    {
+                      symlink (g_file_get_path (current_data_dir),
```
ramcq: this needs error handling too

common/flatpak-run.c
uajain: hmm, currently this patch `break;` the loop as soon as we rename the previous_data_dir and symlinks it (successfully). So it does not makes sense to check the current dir at the start, maybe?

common/flatpak-dir.c
```diff
@@ -6990,6 +7004,19 @@ flatpak_dir_deploy (FlatpakDir          *self,
...
+  if (old_deploy_data)
+    {
+      g_autofree const char **previous_ids = flatpak_deploy_data_get_previous_ids (old_deploy_data);
+      g_variant_builder_add (&metadata_builder, "{s@v}", "previous-ids",
+                             g_variant_new_variant (g_variant_new_strv ((const char * const *) previous_ids, -1)));
+    }
+  else
+    {
+      // hard coded list of dummy ids
+      const char *previous_ids[] = {"com.dummy.com", "com1.dummy1.com1", "com2.dummy2.com2", NULL};
```
pwithnall: What’s the reasoning behind putting in a hard-coded list of dummy IDs?

common/flatpak-dir.c
```diff
@@ -1935,6 +1935,20 @@ flatpak_deploy_data_get_eol_rebase (GVariant *deploy_data)
   return eol;
 }
 
+const char **
+flatpak_deploy_data_get_previous_ids (GVariant *deploy_data)
```
pwithnall: If you added a `gsize *length` argument to this function, you could avoid the `g_strv_length()` call and funny backwards `while` loop below.

common/flatpak-run.c
```diff
@@ -2963,8 +2963,46 @@ flatpak_run_app (const char     *app_ref,
...
+              if (!flatpak_file_rename (previous_data_dir,
+                                        current_data_dir,
+                                        cancellable, &local_error))
+                 {
+                   g_propagate_error (error, g_steal_pointer (&local_error));
+                   return FALSE;
+                 }
+
+               if (!symlink (g_file_get_path (current_data_dir),
+                             g_file_get_path (previous_data_dir)))
```
pwithnall: `g_file_get_path()` returns a newly allocated result, so this will leak.

common/flatpak-run.c
```diff
@@ -2963,8 +2963,46 @@ flatpak_run_app (const char     *app_ref,
...
+
+              if (!flatpak_file_rename (previous_data_dir,
+                                        current_data_dir,
+                                        cancellable, &local_error))
+                 {
+                   g_propagate_error (error, g_steal_pointer (&local_error));
+                   return FALSE;
+                 }
+
+               if (!symlink (g_file_get_path (current_data_dir),
```
pwithnall: Nitpick: Personally, I think it’s clearer to do `if (symlink (…) == 0)` to check for success, rather than `if (!symlink (…))`. The latter reads as “if not symlink”, which suggests failure to me.

common/flatpak-dir.c
uajain: Sorry, I missed this comment.

This is just for testing. This should be already come in the deploy file of the app in the future, I believe. Just that, I shove it directly.

So, if you are testing this PR, and rename one of `~/.var/flatpak/app/any.exisiting.app` to one of these hard-coded IDs and re-run `flatpak run any.exisitng.app`, you should see the renaming and symlinks taking place as intended by this PR.

common/flatpak-dir.c
```diff
@@ -1935,6 +1935,24 @@ flatpak_deploy_data_get_eol_rebase (GVariant *deploy_data)
...
+  g_autoptr(GVariant) metadata = g_variant_get_child_value (deploy_data, 4);
+  g_autoptr(GVariant) previous_ids_v = NULL;
+  const char **previous_ids = NULL;
+
+  previous_ids_v = g_variant_lookup_value (metadata, "previous-ids", NULL);
+  if (previous_ids_v)
+    {
+      previous_ids = g_variant_get_strv (previous_ids_v, NULL);
+      if (length != NULL)
+        *length = g_strv_length ((char **) previous_ids);
```
pwithnall: Just pass `length` directly to `g_variant_get_strv()`:
```
previous_ids = g_variant_get_strv (previous_ids_v, length);
```

common/flatpak-dir.c
```diff
@@ -1935,6 +1935,24 @@ flatpak_deploy_data_get_eol_rebase (GVariant *deploy_data)
...
+    }
```
pwithnall: You should also have:
```
else
  {
    if (length != NULL)
      *length = 0;
  }
```

so that `length` is never left undefined.

uajain: can't we : 
```
else if (length != NULL)
  *length = 0;
```
I think we can, but your code block is more read-able I think :)

common/flatpak-dir.c
```diff
@@ -6990,6 +7009,12 @@ flatpak_dir_deploy (FlatpakDir          *self,
   if (application_runtime)
     g_variant_builder_add (&metadata_builder, "{s@v}", "runtime",
                            g_variant_new_variant (g_variant_new_string (application_runtime)));
+  if (old_deploy_data)
+    {
+      g_autofree const char **previous_ids = flatpak_deploy_data_get_previous_ids (old_deploy_data, NULL);
+      g_variant_builder_add (&metadata_builder, "{s@v}", "previous-ids",
```
ramcq: We should not add the previous-ids field to the new deploy file if it is not present in the old one, so this should be conditional.

common/flatpak-run.c
```diff
@@ -2963,8 +2963,49 @@ flatpak_run_app (const char     *app_ref,
...
   if (app_deploy != NULL)
     {
+      g_autofree const char **previous_ids = NULL;
+      gsize len = 0;
+
       app_files = flatpak_deploy_get_files (app_deploy);
 
+      g_autoptr (GFile) current_data_dir = flatpak_get_data_dir (app_ref_parts[1]);
+      previous_ids = flatpak_deploy_data_get_previous_ids (app_deploy_data, &len);
+      if (previous_ids)
```
ramcq: Because len starts as 0, we can actually skip this if - the loop will be skipped if len is 0. This will reduced the indentation.

PR #108, opened 2018-08-17
Rebase onto upstream master (0.99.4) again

mwleeds:


mwleeds:
The cherry picks applied cleanly, so this just gets us the latest commits from upstream which all seem pretty safe and useful.

manuq:
Yes please :pray:

mwleeds:
Merged after discussion with @ramcq

PR #109, opened 2018-08-20
Rebase onto Flatpak 1.0  🎉

mwleeds:
This is the current state of upstream master (which is two commits past 1.0) plus 8 downstream commits. This gets us a few more commits than what we have already, and notably a fix for T23143.


mwleeds:
Pushed after discussion with @ramcq

PR #110, opened 2018-08-20
Rebase onto Flatpak 1.0 (packaging)

mwleeds:


mwleeds:
When Debian does the packaging of 1.0 we will probably want to rebase onto that, but I don't expect it to be many changes. Technically the flatpak package should depend on ostree 2018.7 not 2018.6 but Endless has all the commits from 2018.7 just without the version bump, so it should be fine as is. The release notes for 1.0 also said to add a weak dependency on p11-kit-server but I think we can wait for the packaging rebase to do that.

mwleeds:
Pushed after discussion with @ramcq

PR #112, opened 2018-09-03
system-helper: set IO class to idle

ramcq:
Our benchmarks show this significantly reduces the interactivity impact of
ongoing Flatpak operations while the user is continuing other tasks on the
system. The effect is very pronounced with the default CFQ scheduler, and in
combination with BFQ, using the idle class improves the worst case to nearly
the same as an unloaded system.

https://phabricator.endlessm.com/T22479


ramcq:
May as well! Submitted https://github.com/flatpak/flatpak/pull/2071 with a copy of the benchmark results.

PR #114, opened 2018-09-18
dir: Support an eos.collection-id metadata key

mwleeds:
Currently flatpak looks for the key "xa.collection-id" in a remote's
metadata to decide whether to update its config with a collection ID.
This commit implements an Endless specific version of it,
eos.collection-id. The purpose of this is so that we can enable a
collection ID on the flathub remote for Endless users without enabling
it for all Flathub users, which is considered risky at the moment.
The Endless versions of flatpak and ostree have patches for various
bugs which other distributions don't have yet.

Eventually once Flathub deploys a collection ID for all users, this
patch should be dropped.

For more discussion see https://phabricator.endlessm.com/T19363#627411

https://phabricator.endlessm.com/T23414

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -11498,6 +11498,7 @@ flatpak_dir_update_remote_configuration_for_state (FlatpakDir         *self,
     "xa.gpg-keys",
     "xa.redirect-url",
     "xa.collection-id",
+    "eos.collection-id", /* Endless specific and temporary */
     NULL
   };
 
@@ -11540,7 +11541,7 @@ flatpak_dir_update_remote_configuration_for_state (FlatpakDir         *self,
                     {
                       if (strcmp (key, "xa.redirect-url") == 0)
                         g_ptr_array_add (updated_params, g_strdup ("url"));
-                      else if (strcmp (key, "xa.collection-id") == 0)
+                      else if (strcmp (key, "xa.collection-id") == 0 || strcmp (key, "eos.collection-id") == 0)
                         g_ptr_array_add (updated_params, g_strdup ("collection-id"));
                       else
                         g_ptr_array_add (updated_params, g_strdup (key));


common/flatpak-dir.c
```diff
@@ -11498,6 +11498,7 @@ flatpak_dir_update_remote_configuration_for_state (FlatpakDir         *self,
     "xa.gpg-keys",
     "xa.redirect-url",
     "xa.collection-id",
+    "eos.collection-id", /* Endless specific and temporary */
```
pwithnall: Support for `eos.collection-id` also needs to be added in `app/flatpak-builtins-repo.c`. You might also need to add support for it in `flatpak_repo_update()` in `common/flatpak-utils.c`. Certainly the comment there (which mentions `xa.collection-id`) should be updated.

manuq:
Will this change affect scripts like `eos-stage-flatpak` ? https://github.com/endlessm/eos-meta/blob/master/eos-tech-support/eos-stage-flatpak#L73

mwleeds:
> Will this change affect scripts like `eos-stage-flatpak` ? https://github.com/endlessm/eos-meta/blob/master/eos-tech-support/eos-stage-flatpak#L73

The logic for updating the collection ID on a remote in `eos-stage-flatpak`/`eos-stage-ostree` is still valid regardless of what approach we go with here.

mwleeds:
> Would it be helpful if the key were defined to only be processed by sufficiently new libostree/flatpak versions? That's implicitly what we're doing here right?

Currently in either approach, here or in https://github.com/flatpak/flatpak/pull/2119, it's written so the key is only processed when flatpak is new enough, under the assumption that libostree is fairly updated as well (it's [guaranteed to be at least 2018.7](https://github.com/flatpak/flatpak/blob/master/configure.ac#L35)). I guess you're suggesting to change the upstream PR so that it only does anything if `OSTREE_META_KEY_DEPLOY_COLLECTION_ID` is defined, which seems like a good idea. The only reason I can think of not to do that is if some distribution were packaging ostree 2018.7 with bugfix patches applied but without the patch defining that key. But realistically Endless is probably the only one using this in the near term future.

mwleeds:
[The alternative approach](https://github.com/flatpak/flatpak/pull/2119) was merged. Closing

PR #115, opened 2018-09-20
T15371: allow installing via system helper without copying

uajain:
Wip branch for shooting double disk issue.

PR opened for more concrete discussion(implementation-wise).

https://phabricator.endlessm.com/T15371


system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +205,123 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+  if (ongoing_pulls == NULL)
+    ongoing_pulls = g_ptr_array_new ();
+
+  msg = g_dbus_message_new_method_call ("org.freedesktop.DBus",
+                                        "/org/freedesktop/DBus",
+                                        "org.freedesktop.DBus",
+                                        "GetConnectionCredentials");
+  g_dbus_message_set_body (msg, g_variant_new ("(s)", sender));
+
+  reply = g_dbus_connection_send_message_with_reply_sync (connection, msg,
```
pwithnall: It would be significantly simpler to call `g_dbus_connection_call_sync()` instead.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +205,123 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+            }
+        }
+    }
+
+  for (guint i = 0; i < ongoing_pulls->len; i++)
+    {
+      FlatpakSystemHelperOngoingPull *pull_tmp;
+
+      pull_tmp = g_ptr_array_index (ongoing_pulls, i);
+      if (pull_tmp->uid = uid)
```
pwithnall: `==`

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +205,123 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    }
+
+  if (uid_pull_count > MAX_PULLS_PER_USER)
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                             "Maximum number of pulls per user exceeded for UID: %u", uid);
+      return TRUE;
+    }
+
+  cache_tmpdir = g_strdup ("flatpak-cache-XXXXXX");
```
pwithnall: No need for the `g_strdup()` here.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +205,123 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                             "Cannot create directory for bindfs mount at: %s", user_folder);
+      return TRUE;
+    }
+
+  new_pull = system_helper_ongoing_pull_new (connection, uid, arg_ref, optional_commit, user_folder);
+  g_ptr_array_add (ongoing_pulls, new_pull);
+
+  if (g_find_program_in_path ("bindfs") != NULL)
```
pwithnall: This will leak the program’s path.

system-helper/flatpak-system-helper.c
```diff
@@ -1245,7 +1426,8 @@ flatpak_authorize_method_handler (GDBusInterfaceSkeleton *interface,
            g_strcmp0 (method_name, "PruneLocalRepo") == 0 ||
            g_strcmp0 (method_name, "EnsureRepo") == 0 ||
            g_strcmp0 (method_name, "RunTriggers") == 0 ||
-           g_strcmp0 (method_name, "UpdateSummary") == 0)
+           g_strcmp0 (method_name, "UpdateSummary") == 0 ||
+	   g_strcmp0 (method_name, "CreatePullCacheDir") == 0)
```
pwithnall: Nitpick: Indentation seems wrong.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +205,123 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+
+  system = dir_get_system (arg_installation, &error);
+  if (system == NULL)
+    {
+      g_dbus_method_invocation_return_gerror (invocation, error);
+      return TRUE;
+    }
+
+  if (ongoing_pulls == NULL)
+    ongoing_pulls = g_ptr_array_new ();
```
pwithnall: This never gets freed, which will make valgrinding the system helper harder, by showing up in the valgrind output as a kind of false positive.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +205,123 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    {
+      FlatpakSystemHelperOngoingPull *pull_tmp;
+
+      pull_tmp = g_ptr_array_index (ongoing_pulls, i);
+      if (pull_tmp->uid = uid)
+        {
+          uid_pull_count++;
+          /*
+          * If there is a already a partial pull available, return that.
+           */
```
pwithnall: Is this a comment describing the code in its current state, or describing something which is still to be implemented? If the latter, I suggest prefixing it with `TODO: ` so that it’s clear to you (via `git grep`) what still needs to be done; and it’s clear to any reviewers too.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +206,137 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+  g_autoptr(GVariant) value = NULL;
+  uid_t uid;
+
+  credentials = g_dbus_connection_call_sync (connection,
+                                             "org.freedesktop.DBus",
+                                             "/org/freedesktop/DBus",
+                                             "org.freedesktop.DBus",
+                                             "GetConnectionCredentials",
+                                             g_variant_new ("(s)", sender),
+                                             NULL, G_DBUS_CALL_FLAGS_NONE, G_MAXINT, NULL, error);
```
wjt: Pass `G_VARIANT_TYPE ("(a{sv})")` for the (currently-`NULL`) `reply_type` parameter here.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +206,137 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+
+  if (credentials == NULL)
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                             "Error calling GetConnectionCredentials over D-Bus");
+      return TRUE;
```
wjt: This doesn't seem right: `error` has been filled in but the caller thinks the connection is Unix user ID `FALSE` (ie `1`). Since all values of `uid_t` are valid UIDs, maybe change the signature to return `gboolean` and take a `uid_t *` out parameter?

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +206,137 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    }
+
+  g_variant_get (credentials, "(a{sv})", &iter);
```
wjt: Assuming the function has `gboolean` return type and a `uid_t *uid_out` parameter, this loop can be replaced with something like:

```
g_autoptr(GVariant) dict = g_variant_get_child_value (credentials, 0);
g_autoptr(GVariant) unixUserId = g_variant_lookup_value (dict, "UnixUserID", G_VARIANT_TYPE_UINT32);

if (unixUserId == NULL) { set error; return FALSE; }
*uid_out = g_variant_get_uint32 (unixUserId);
return TRUE;
```

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +206,137 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+        {
+          uid_pull_count++;
+          // TODO: If there is a already a partial pull available, return that.
+        }
+    }
+
+  uid = get_connection_uid (invocation, &error);
+  if (uid_pull_count > MAX_PULLS_PER_USER)
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
```
wjt: `G_DBUS_ERROR_LIMITS_EXCEEDED`?

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +206,137 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+  g_ptr_array_add (ongoing_pulls, new_pull);
+
+  bindfs_path = g_find_program_in_path ("bindfs");
+  if (bindfs_path != NULL)
+    {
+      g_autofree gchar *arg_user = NULL;
+
+      arg_user = g_strdup_printf ("--force-user=%"G_GUINT32_FORMAT, uid);
+      new_pull->bindfs = g_subprocess_new (G_SUBPROCESS_FLAGS_NONE,
+                                           NULL,
```
wjt: Probably want to pass `&error` here so you can check if it succeeded?

system-helper/flatpak-system-helper.c
```diff
@@ -172,6 +371,27 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+      pull_tmp = g_ptr_array_index (ongoing_pulls, i);
+      if (g_strcmp0(pull_tmp->user_folder_name, user_folder) == 0)
+        {
+          ongoing_pull = pull_tmp;
+
+	  if (umount (user_folder) == -1)
+            g_warning ("flatpak-cache: Could not unmount directory %s\n", user_folder);
+
+          if (ongoing_pull->bindfs)
+            g_subprocess_force_exit (pull_tmp->bindfs);
```
wjt: You may want to `g_subprocess_wait()` here because `force_exit()` doesn't wait for the process to die before returinng.

system-helper/flatpak-system-helper.c
```diff
@@ -46,11 +46,63 @@ static gboolean no_idle_exit = FALSE;
...
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitDetails, g_object_unref)
 G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitSubject, g_object_unref)
 
+struct _FlatpakSystemHelperOngoingPull
+{
+  FlatpakSystemHelper *object;
+  GDBusMethodInvocation *invocation;
+
+  const gchar *commit_id;
```
wjt: Who owns these strings?

system-helper/flatpak-system-helper.c
```diff
@@ -46,11 +46,63 @@ static gboolean no_idle_exit = FALSE;
...
+  gchar *root_folder_name;
+  gchar *user_folder_name;
+
+  GUnixMountMonitor *mount_monitor;
+  gulong mount_monitor_id;
+  GSubprocess *bindfs;
+};
+
+static void
+system_helper_ongoing_pull_free (FlatpakSystemHelperOngoingPull *pull)
```
wjt: A bit surprising to see a _free  function that doesn't actually free its argument.

system-helper/flatpak-system-helper.c
```diff
@@ -193,6 +232,160 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+
+static void
+mounts_changed (GUnixMountMonitor *monitor, gpointer user_data)
+{
+  FlatpakSystemHelperOngoingPull *pull = (FlatpakSystemHelperOngoingPull *) user_data;
+  g_autoptr(GList) mounts = NULL;
+  GList *l = NULL;
+
+  mounts = g_unix_mounts_get (&pull->current_timestamp);
+  for (l = g_list_reverse (mounts); l != NULL; l = l->next)
```
wjt: Why reverse this list before scanning it?

system-helper/flatpak-system-helper.c
```diff
@@ -193,6 +232,160 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    }
+
+  if (ongoing_pulls == NULL)
+    ongoing_pulls = g_ptr_array_new ();
+
+  for (guint i = 0; i < ongoing_pulls->len; i++)
+    {
+      FlatpakSystemHelperOngoingPull *pull_tmp;
+      pull_tmp = g_ptr_array_index (ongoing_pulls, i);
+      if (pull_tmp->uid == uid)
```
wjt: `uid` isn't defined until after this loop

system-helper/flatpak-system-helper.c
```diff
@@ -193,6 +232,160 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                             "Cannot create directory for bindfs mount at: %s", user_folder);
+      return TRUE;
+    }
+
+  new_pull = system_helper_ongoing_pull_new (object, invocation, uid, arg_ref, optional_commit, user_folder);
+  g_ptr_array_add (ongoing_pulls, new_pull);
+
+  bindfs_path = g_find_program_in_path ("bindfs");
```
wjt: This should be checked much earlier in this function.

system-helper/flatpak-system-helper.c
```diff
@@ -224,6 +418,44 @@ handle_deploy (FlatpakSystemHelper   *object,
       return TRUE;
     }
 
+  for (guint i = 0; i < ongoing_pulls->len; i++)
+    {
+      FlatpakSystemHelperOngoingPull *pull_tmp;
+      g_autofree gchar *user_folder = NULL;
+
+      user_folder = g_file_get_path (g_file_get_parent (path));
```
wjt: Why not do this outside the loop? `path` doesn't change between iterations.

Consider factoring the loop out to a `get_ongoing_pull_by_dir ()` function that returns a possibly-NULL `FlatpakSystemHelperOngoingPull`

system-helper/flatpak-system-helper.c
```diff
@@ -224,6 +418,44 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+              g_set_error (&error, G_IO_ERROR, g_io_error_from_errno (errsv),
+                           "Unmounting :%s failed: %s\n", user_folder, g_strerror (errsv));
+              g_dbus_method_invocation_return_gerror (invocation, error);
+              return TRUE;
+            }
+
+	  if (ongoing_pull->bindfs)
+            {
+              g_subprocess_wait (ongoing_pull->bindfs, NULL, NULL);
+              g_subprocess_force_exit (ongoing_pull->bindfs);
```
wjt: `wait()` blocks until the subprocess dies, at which point killing it is not necessary. We only need to kill it in the case where `umount` fails because it's in use, and we need to do that before waiting for it to terminate.

system-helper/flatpak-system-helper.c
wjt: Also, `g_file_get_parent()` returns a new ref, which is leaked here

system-helper/flatpak-system-helper.c
```diff
@@ -224,6 +418,44 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+            }
+
+	  g_autofree gchar *new_path = g_build_filename (ongoing_pull->root_folder_name, g_file_get_basename (path), NULL);
+          g_clear_object(&path);
+          path = g_file_new_for_path (new_path);
+          arg_repo_path = g_steal_pointer (&new_path);
```
wjt: This pointer is leaked.

common/flatpak-dir.c
```diff
@@ -1164,6 +1164,30 @@ flatpak_dir_system_helper_call (FlatpakDir   *self,
                                       error);
 }
 
+static gboolean
+flatpak_dir_system_helper_call_create_system_child_repo (FlatpakDir    *self,
```
wjt: This function name should be consistent with the D-Bus method. Does it create a repo, or just a directory within the repo?

common/flatpak-dir.c
```diff
@@ -7602,23 +7635,33 @@ flatpak_dir_install (FlatpakDir          *self,
           if (child_repo == NULL)
             return FALSE;
 
-          flatpak_flags |= FLATPAK_PULL_FLAGS_SIDELOAD_EXTRA_DATA;
+          bindfs_repo_file = g_file_new_for_path (bindfs_repo_path);
+          bindfs_repo = flatpak_dir_create_child_repo (self, bindfs_repo_file, &bindfs_repo_lock, NULL, error);
+          if (bindfs_repo == NULL)
+            {
+              g_warning ("Cannot create child repo on given path");
```
wjt:
```c
g_prefix_error (error, _("Cannot create child repo at %s: "), bindfs_repo_path);
```

might be better?

common/flatpak-dir.c
```diff
@@ -7594,6 +7618,15 @@ flatpak_dir_install (FlatpakDir          *self,
...
       else
         {
+          g_autofree gchar *bindfs_repo_path = NULL;
+          g_autoptr(GFile) bindfs_repo_file = NULL;
+          g_auto(GLnxLockFile) bindfs_repo_lock = { 0, };
+          g_autoptr(OstreeRepo) bindfs_repo = NULL;
+
+          flatpak_dir_system_helper_call_create_system_child_repo (self, ref, opt_commit,
+                                                                   installation ? installation : "",
+                                                                   &bindfs_repo_path, cancellable, error);
```
wjt: Any error returned here is ignored. Then we unconditionally call `flatpak_dir_create_system_child_repo` below – but I think we should only do that if this call fails with `G_DBUS_ERROR_UNKNOWN_METHOD`.

common/flatpak-dir.c
```diff
@@ -7602,23 +7635,33 @@ flatpak_dir_install (FlatpakDir          *self,
           if (child_repo == NULL)
             return FALSE;
 
-          flatpak_flags |= FLATPAK_PULL_FLAGS_SIDELOAD_EXTRA_DATA;
+          bindfs_repo_file = g_file_new_for_path (bindfs_repo_path);
```
wjt: this path may be NULL if the call above failed

system-helper/flatpak-system-helper.c
uajain: This is one of those moments: "What was I thinking" !

system-helper/flatpak-system-helper.c
uajain: hmm, I was getting the mounts without the timestamp in an earlier version of the code. Therefore I used to reverse the list as the bindfs mount was getting found at the last. Less no. of loops* but anyway, with using timestamp it can be dropped. Thanks for noticing.

wjt: Well, reversing the list requires iterating it in full..

system-helper/flatpak-system-helper.c
```diff
@@ -46,11 +46,63 @@ static gboolean no_idle_exit = FALSE;
...
+  if (pull->mount_monitor_id)
+    {
+      g_signal_handler_disconnect (pull->mount_monitor, pull->mount_monitor_id);
+      pull->mount_monitor_id = 0;
+    }
+  g_object_unref (pull->mount_monitor);
+
+  g_ptr_array_remove (ongoing_pulls, pull);
+  g_ptr_array_unref (ongoing_pulls);
+  if (ongoing_pulls->len == 0)
```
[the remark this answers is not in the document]
uajain: Question:
... And live where?

Sure, the unreffing should occur inside the `if` block (my approach here is that, the last ongoing_pull should `unref` the `ongoing_pulls` and set it NULL. If there are other ongoing pulls, (ongoing_pulls->len > 0), we shouldn't.

Also, after reading the `g_ptr_array_free` documentation, I think having a `g_ptr_array_ref/unref` pair approach is what is desirable here. Thoughts?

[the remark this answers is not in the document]
wjt: I suggest allocating the array at the start of main(), and unreffing it at the end, rather than trying to arrange for it to be `NULL` when it would otherwise be empty.

Thinking about this a bit more, and reading the code: the system helper currently exits 10 minutes after it is launched. If an unprivileged pull takes longer than 10 minutes, the helper will exit with the bind still set up. When the app comes to call Deploy(), the helper will be relaunched but will have no idea that it created the bindfs for this user. I think we need to keep the daemon alive for as long as there are any active bindfs mounts.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +295,157 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+
+  *uid_out = g_variant_get_uint32 (unixUserId);
+  return TRUE;
+}
+
+static gboolean
+handle_create_pull_cache_dir (FlatpakSystemHelper   *object,
+                              GDBusMethodInvocation *invocation,
+                              const gchar           *arg_ref,
+                              const gchar           *optional_commit,
```
wjt: The ref and commit arguments appear to be unused. Why are they here?

system-helper/flatpak-system-helper.c
```diff
@@ -25,32 +25,186 @@
...
+}
+
+static void
+mounts_changed (GUnixMountMonitor *monitor, gpointer user_data)
+{
+  FlatpakSystemHelperOngoingPull *pull = (FlatpakSystemHelperOngoingPull *) user_data;
+  g_autoptr(GList) mounts = NULL;
+  GList *l = NULL;
+
+  mounts = g_unix_mounts_get (&pull->current_timestamp);
```
wjt: The return value of this function is "transfer full" but this function only cleans up the GList container. It will leak all the `GUnixMountEntry *` values in the list.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +295,157 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+      return TRUE;
+    }
+
+  flatpak_dir = g_file_get_path (flatpak_dir_get_path (system));
+  root_folder = g_mkdtemp_full (g_build_filename (flatpak_dir, "repo", "tmp", "flatpak-cache-XXXXXX", NULL), 0755);
+  user_folder = g_strdup_printf ("%s-bindfs", root_folder);
+  if (g_mkdir_with_parents (user_folder, 0755) == -1)
+    {
+      g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+                                             "Cannot create directory for bindfs mount at: %s", user_folder);
```
wjt: Would be nice to include `g_strerror (errno)` in here. You may be interested in `glnx_throw_errno_prefix (&error, "Can't create directory for bindfs mount at %s", user_folder)`, here and the other places where `errno` is used. (I'd forgotten that Flatpak depends on libglnx which has these nice helper functions to turn `errno` into a GError.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +295,157 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+  new_pull = system_helper_ongoing_pull_new (object, invocation, uid, arg_ref, optional_commit, user_folder);
+  g_ptr_array_add (ongoing_pulls, new_pull);
+
+  bindfs_path = g_find_program_in_path ("bindfs");
+  if (bindfs_path != NULL)
+    {
+      g_autofree gchar *arg_user = NULL;
+      g_autoptr(GFile) user_folder_file = NULL;
+
+      user_folder_file = g_file_new_for_path (user_folder);
```
wjt: This is unused.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +295,157 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+  return TRUE;
+}
+
+static FlatpakSystemHelperOngoingPull *
+get_ongoing_pull_by_dir (const gchar *arg_repo_path)
+{
+  FlatpakSystemHelperOngoingPull *pull;
+  g_autoptr(GFile) repo_file = g_file_new_for_path (arg_repo_path);
+  g_autoptr(GFile) parent = g_file_get_parent (repo_file);
+  g_autofree gchar *user_folder = g_file_get_path (parent);
```
wjt:
```suggestion
  g_autofree gchar *user_folder = g_path_get_dirname (arg_repo_path);
```

wjt: (and remove `repo_file` and `parent`)

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +295,157 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+                                           "--create-for-group=root",
+                                           root_folder, user_folder, NULL);
+      if (new_pull->bindfs == NULL)
+        {
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_SPAWN_FAILED,
+                                                 "Error spawning bindfs for user:%"G_GUINT32_FORMAT" at: %s", uid, user_folder);
+          return TRUE;
+        }
+
+      new_pull->root_folder_name = g_strdup (root_folder);
```
wjt:
```suggestion
      new_pull->root_folder_name = g_steal_pointer (&root_folder);
```

system-helper/flatpak-system-helper.c
uajain: I have gone all the way to check `bindfs` in the client code. If `bindfs` is not found, then it fallbacks on the existing code path for the pull.

These bit is addressed in https://github.com/endlessm/flatpak/pull/115/commits/b55725c01a0eb1a016dde7e362052c02506d20fc

common/flatpak-dir.c
uajain: This just creates a directory, yes. I should probably think of something entirely different which is more apt. Adding to my ToDo list.

common/flatpak-dir.c
uajain: These bit is addressed in b55725c
Can you take a look at that first and reply?

common/flatpak-dir.c
uajain: hmm, As per my changes in  b55725c , I still think to display a warning here. The idea is, if we fail on bindfs code-path, we just don't want to error out; instead fallback on existing code-path and display a warning (that there was something wrong with bindfs code path). Does it make sense?

common/flatpak-dir.c
uajain: These bit is addressed in b55725c

system-helper/flatpak-system-helper.c
```diff
@@ -25,32 +25,186 @@
...
+  FlatpakSystemHelperOngoingPull *pull = (FlatpakSystemHelperOngoingPull *) user_data;
+  g_autoptr(GList) mounts = NULL;
+  GList *l = NULL;
+
+  mounts = g_unix_mounts_get (&pull->current_timestamp);
+  for (l = mounts; l != NULL; l = l->next)
+    {
+      if (g_strcmp0 (pull->user_folder_name, g_unix_mount_get_mount_path (l->data)) == 0)
+        {
+          if (pull->mount_monitor_id != 0)
```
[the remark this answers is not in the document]
uajain: Right. This is copy/pasting fiasco.

common/flatpak-dir.c
wjt: I think the code change there makes sense, modulo my comment on that commit.

system-helper/flatpak-system-helper.c
uajain: Addressed this in a172ca3f91aa5a003550a90de4dba0c6ab6b2441 with few others(which I have marked resolved)

system-helper/flatpak-system-helper.c
uajain: Addressed this in a172ca3f91aa5a003550a90de4dba0c6ab6b2441 with few others(which I have marked resolved)

system-helper/flatpak-system-helper.c
```diff
@@ -172,6 +479,41 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+      g_autofree gchar *bindfs_repo_path = NULL;
+      g_autofree gchar *real_repo_path = NULL;
+
+      if (umount (ongoing_pull->user_folder_name) == -1)
+        {
+          int errsv = errno;
+          g_set_error (&error, G_IO_ERROR, g_io_error_from_errno (errsv),
+                       "Unmounting :%s failed: %s\n", ongoing_pull->user_folder_name, g_strerror (errsv));
+          g_dbus_method_invocation_return_gerror (invocation, error);
+          return TRUE;
```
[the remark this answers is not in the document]
uajain: This bit is addressed in https://github.com/endlessm/flatpak/pull/115/commits/e112d51951dc17070ac530b83a922c078f13bf3b

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
uajain: I have tried to address this bit in https://github.com/endlessm/flatpak/pull/115/commits/15b016e99e838a6d43d64b50212a6e36e482243f

Thanks for the suggestion.

system-helper/flatpak-system-helper.c
uajain: Also, this(`g_find_program_in_path`) needs to be queried for `bindfs` path which is used for the spawing the subprocess below. Maybe get rid of the `if` condition here (as it's already being checked on the client side now)?

system-helper/flatpak-system-helper.c
```diff
@@ -248,6 +245,12 @@ unref_skeleton_in_timeout (void)
 static gboolean
 idle_timeout_cb (gpointer user_data)
 {
+  if (ongoing_pulls->len != 0)
+    return G_SOURCE_CONTINUE;
+
+  g_ptr_array_unref (ongoing_pulls);
+  ongoing_pulls = NULL;
```
wjt: No, I think this should go in `main()`, after `g_main_loop_run()`

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +295,157 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+    }
+
+  for (guint i = 0; i < ongoing_pulls->len; i++)
+    {
+      FlatpakSystemHelperOngoingPull *pull_tmp;
+      pull_tmp = g_ptr_array_index (ongoing_pulls, i);
+      if (pull_tmp->uid == uid)
+        {
+          uid_pull_count++;
+          // TODO: If there is a already a partial pull available, return that.
```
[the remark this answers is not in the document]
uajain: Yes. That was the idea.
But as decided in the daily, I will tackle this bit in later patch. Therefore, dissecting these args  out of the system-helper for now. 
Introducing it in the later patch will make more sense. Resolving this and related comments about these args.

common/flatpak-dir.c
```diff
@@ -7594,31 +7613,76 @@ flatpak_dir_install (FlatpakDir          *self,
         }
       else
         {
+          /* If we find bindfs on the system, we try to make a pull using the bindfs mount otherwise fallback.
+             Also, we need a separate OSTree pointer to bindfs repo, local to this block.
+             Reason being that if we assign the bindfs repo to `child_repo` which has life outside
+             this block, while deploying, it will lead to open FDs and unmounting the bindfs mount
+             will fail with EBUSY (inside system-helper's handle_deploy()). */
```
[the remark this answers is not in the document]
uajain: hmm, this is interesting. Earlier I have tried out :
``` 
child_repo = g_steal_pointer (&bindfs_repo)  
...
g_object_unref (child_repo) // to close it
``` 
I was error-ing out for cleanup code path:
```
#0  0x00007ffff77263c5 in  () at /lib64/libglib-2.0.so.0
#1  0x00007ffff77276fd in g_logv () at /lib64/libglib-2.0.so.0
#2  0x00007ffff77278d3 in g_log () at /lib64/libglib-2.0.so.0
#3  0x000000000045f544 in glib_autoptr_cleanup_OstreeRepo (_ptr=<synthetic pointer>)
    at /usr/include/ostree-1/ostree-autocleanups.h:55
#4  0x000000000045f544 in flatpak_dir_install
    (self=0x4f1800, no_pull=<optimized out>, no_deploy=0, no_static_deltas=<optimized out>, reinstall=0, state=state@entry=0x501b10, ref=0x630790 "app/org.gnome.Photos/x86_64/stable", opt_commit=0x503970 "a89502fbf33367163e5e09d785284063658577babab1705d088dca5dea3c9e14", opt_subpaths=<optimized out>, progress=0x7fffe400dc40, cancellable=0x0, error=0x7fffffffd3e8)
    at common/flatpak-dir.c:7546
#5  0x0000000000470fa6 in flatpak_transaction_run
    (self=self@entry=0x4fc930, cancellable=cancellable@entry=0x0, error=error@entry=0x7fffffffd438)
    at common/flatpak-transaction.c:2384

```
But `g_clear_pointer (&child_repo)` doesn't give me any error. I wonder why? :thinking: 

Reducing the scope might work but it will introduce changes in the upper `else if` blocks and need to be declared in those blocks separately.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +291,148 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+
+static gboolean
+handle_create_pull_cache_dir (FlatpakSystemHelper   *object,
+                              GDBusMethodInvocation *invocation,
+                              const gchar           *arg_installation)
+{
+  GDBusConnection *connection = g_dbus_method_invocation_get_connection (invocation);
+  const gchar *sender = g_dbus_method_invocation_get_sender (invocation);
+  FlatpakSystemHelperOngoingPull *new_pull;
+  FlatpakSystemHelperOngoingPull *pull_tmp;
```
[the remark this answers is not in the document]
uajain: It is used below [inside the loop.](https://github.com/endlessm/flatpak/pull/115/files#diff-9885ca1a8cc5c8690061c7fed45ddf20R359)

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +57,147 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+name_vanished_callback (GDBusConnection *connection, const gchar *name, gpointer user_data)
+{
+  system_helper_ongoing_pull_free ((FlatpakSystemHelperOngoingPull *) user_data);
+}
+
+static void
+mounts_changed (GUnixMountMonitor *monitor, gpointer user_data)
+{
+  FlatpakSystemHelperOngoingPull *pull = user_data;
+  g_autoptr(GList) mounts = NULL;
```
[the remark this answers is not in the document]
uajain: Turns out, we need to wait for glib version `2.56` in order to use it.
But good to know about this for future references!

common/flatpak-dir.c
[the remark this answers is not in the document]
wjt: Why? Because calling `g_object_unref (child_repo)` does not set `child_repo = NULL`, so when you leave the scope, `g_object_unref (child_repo)` is called again ⇒ double free.

common/flatpak-dir.c
```diff
@@ -7594,13 +7628,42 @@ flatpak_dir_install (FlatpakDir          *self,
...
+            {
+              if (!g_error_matches (*error, G_DBUS_ERROR, G_DBUS_ERROR_UNKNOWN_METHOD))
+                g_warning ("Cannot create pull cache dir in repo/tmp: %s", (*error)->message);
+            }
+          else
+            {
+              g_autoptr(GFile) repo_tmp_cache_dir_gfile = NULL;
+
+              repo_tmp_cache_dir_gfile = g_file_new_for_path (repo_tmp_cache_dir);
+              child_repo = flatpak_dir_create_child_repo (self, repo_tmp_cache_dir_gfile, &child_repo_lock, NULL, error);
```
wjt: Do we need to arrange for CancelPull to be called if this fails?

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +301,177 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+      if (pull_tmp->bindfs == NULL)
+        {
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_SPAWN_FAILED,
+                                                 "Error spawning bindfs for user:%"G_GUINT32_FORMAT" at: %s", uid, user_dir);
+          return TRUE;
+        }
+
+      pull_tmp->root_dir_name = g_steal_pointer (&root_dir);
+
+      new_pull = g_steal_pointer (&pull_tmp);
```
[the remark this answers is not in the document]
wjt: Oh no, it can't be, because it is used for the `wait_check` call below.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +301,177 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+      G_LOCK (ongoing_pulls);
+      g_ptr_array_add (ongoing_pulls, new_pull);
+      G_UNLOCK (ongoing_pulls);
+
+      g_subprocess_wait_check_async (new_pull->bindfs, NULL, subprocess_wait_check_cb, new_pull);
```
wjt: To solve the problem where the callback may be called after `new_pull`:

* Add a `GCancellable` to the OngoingPull struct
* Cancel it at the very start of system_helper_ongoing_pull_free
* Pass it here
* In `subprocess_wait_check_cb`, check if `g_error_matches (error, G_IO_ERROR, G_IO_ERROR_CANCELLED)` and if so take no action.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +301,177 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+        }
+
+      G_LOCK (ongoing_pulls);
+      for (guint i = 0; i < ongoing_pulls->len; i++)
+        {
+          pull_tmp = g_ptr_array_index (ongoing_pulls, i);
+          if (pull_tmp->uid == uid)
+            {
+              uid_pull_count++;
+              // TODO: If there is a already a partial pull available, return that.
```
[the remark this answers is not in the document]
wjt: Let's do option 3: don't worry about breaking the signature of this method.

common/flatpak-dir.c
```diff
@@ -7594,31 +7628,86 @@ flatpak_dir_install (FlatpakDir          *self,
...
+              g_autoptr(GFile) repo_tmp_cache_dir_gfile = NULL;
+
+              repo_tmp_cache_dir_gfile = g_file_new_for_path (repo_tmp_cache_dir);
+              child_repo = flatpak_dir_create_child_repo (self, repo_tmp_cache_dir_gfile, &child_repo_lock, NULL, &local_error);
+              if (child_repo == NULL)
+                {
+                 /* Explicitly clearing the child_repo pointer closes any open FDs on the child_repo.
+                    This prevents EBUSY error failure during unmounting. */
+                  glnx_release_lock_file (&child_repo_lock);
+                  g_clear_object (&child_repo);
```
wjt: By definition, this line cannot be needed, since it's within `if (child_repo == NULL) { ... }`.

common/flatpak-dir.c
```diff
@@ -7594,31 +7628,86 @@ flatpak_dir_install (FlatpakDir          *self,
         }
       else
         {
+          /* If the system-helper can find bindfs on the system, we try to make a pull using the bindfs by
+             mounting repo/tmp/flatpak-cache-XXXXXX directory. If this operation is unsupported or failed,
+             the pull is then tried with the fallback codepath. */
+
+          g_autofree gchar *repo_tmp_cache_dir = NULL;
+          g_autoptr(OstreeRepo) child_repo = NULL;
```
wjt: This shadows the definition at the top-level scope – very misleading.

uajain:
Thanks for taking a look @ramcq, @wjt 
Addressed your comments in the latest fixup c3df571

mwleeds:
Some feedback on the commit messages:

- For "system-helper: Add CreatePullCachedir D-Bus method", perhaps it's worth explaining exactly what "the double disk space situation during the deploy" is. Ideally you might also explain why bindfs is being used.
- For "common: Use system-helper's CreatePullCacheDir D-Bus method during install", I think it's worth explaining why you're changing the way pulls work, even if you'll repeat some of what you said in the above commit message.
- It probably goes without saying that the phabricator links shouldn't be there when this is submitted upstream. Up to you whether you want to keep them downstream until the next rebase.

mwleeds:
`bindfs` isn't part of the base ostree right? Is the plan to add it as a suggested dependency in our downstream packaging?

mwleeds:
And as discussed in the desktop call, the commit message should link to https://github.com/ostreedev/ostree/issues/1723 (and ideally that issue should be [moved to](https://blog.github.com/changelog/2018-10-30-issue-transfer/) the flatpak repo but you'd have to convince an admin of the ostree repo to do that for you).

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +311,188 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_LIMITS_EXCEEDED,
+                                                 "Maximum number of pulls per user exceeded for UID: %u", uid);
+          return TRUE;
+        }
+
+      // TODO: If there is a already partial pull available, return that.
+
+      flatpak_dir = g_file_get_path (flatpak_dir_get_path (system));
+      root_dir = g_mkdtemp_full (g_build_filename (flatpak_dir, "repo", "tmp", "flatpak-cache-XXXXXX", NULL), 0755);
+      user_dir = g_strdup_printf ("%s-bindfs", root_dir);
```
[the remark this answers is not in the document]
uajain: Yes, one can achieve that by using `--force-group=` but one minor issue with that, I didn't find anything like: `get_gid_for_uid(uid) ; but it works out even if `--force-group` is omitted.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +311,188 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+      if (g_mkdir_with_parents (user_dir, 0755) == -1)
+        {
+          glnx_throw_errno_prefix (&error, "Can't create directory for bindfs mount at %s", user_dir);
+          g_dbus_method_invocation_return_gerror (invocation, error);
+          return TRUE;
+        }
+
+      pull_tmp = system_helper_ongoing_pull_new (object, invocation, uid, user_dir);
+
+      arg_user = g_strdup_printf ("--force-user=%"G_GUINT32_FORMAT, uid);
```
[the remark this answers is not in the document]
uajain: This creates a bind fuse mount which let's a user to write to a root-owned directory.

Therefore, the user sees the mirror directory (`$tmpdir-bindfs`) as if it's owned by it`and writes to it. But the actual file creation is being happening in the root directory (`$tmpdir`) 

In the man pages, `--force-user` is under `FILE OWNERSHIP` and `--create-for-user` goes under `FILE CREATION` section; And when the mirrored directory is unmounted, the files are (obviously) owned by root. Does it answer you question?

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,162 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+}
+
+static gboolean
+unsuccessful_mount_timeout (gpointer user_data)
+{
+  g_autoptr(FlatpakSystemHelperOngoingPull) pull = user_data;
+  g_autoptr(GError) error = NULL;
+
+  g_set_error (&error, G_DBUS_ERROR, G_DBUS_ERROR_FAILED,
+               "Failed to mount %s on %s with bindfs", pull->user_dir_name, pull->root_dir_name);
```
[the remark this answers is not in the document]
uajain: May I ask how to trigger the error intentionally?
i tried to reproduce it by letting the timeout fire; and :
```
uajain@endless:~/flatpak$ flatpak install --system flathub org.gnome.Todo
Required runtime for org.gnome.Todo/x86_64/stable (runtime/org.gnome.Platform/x86_64/3.28) found in remote flathub
Do you want to install it? [y/n]: y
Installing in system:
org.gnome.Platform/x86_64/3.28             flathub 6d1d0ebbd724
org.freedesktop.Platform.ffmpeg/x86_64/1.6 flathub d757f762489e
org.gnome.Platform.Locale/x86_64/3.28      flathub 2823e3d81b74
org.gnome.Todo/x86_64/stable               flathub ffd54b66649f
  permissions: ipc, network, wayland, x11
  file access: home, xdg-run/dconf, ~/.config/dconf:ro
  dbus access: ca.desrt.dconf, org.gnome.OnlineAccounts, org.gnome.evolution.dataserver.AddressBook9, org.gnome.evolution.dataserver.Calendar7, org.gnome.evolution.dataserver.Sources5, org.gnome.evolution.dataserver.Subprocess.Backend.*
  system dbus access: org.freedesktop.login1
org.gnome.Todo.Locale/x86_64/stable        flathub 1c1e8e1e15ba
Is this ok [y/n]: y
Installing: org.gnome.Platform/x86_64/3.28 from flathub
flatpak install: Cannot create pull cache dir in repo/tmp: GDBus.Error:org.freedesktop.DBus.Error.Failed: Failed to mount /usr/var/lib/flatpak/repo/tmp/flatpak-cache-0O0ESZ-bindfs on /usr/var/lib/flatpak/repo/tmp/flatpak-cache-0O0ESZ with bindfs
[######              ] Downloading files: 5024/21468 51.8 MB (2.4 MB/s)
```
Error string looks OK to me, throws error and fallbacks to exisitng codepath for pull.

system-helper/flatpak-system-helper.c
```diff
@@ -51,6 +59,162 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (AutoPolkitAuthorizationResult, g_object_unref)
...
+  GSubprocess *bindfs;
+} FlatpakSystemHelperOngoingPull;
+
+static void
+system_helper_ongoing_pull_free (FlatpakSystemHelperOngoingPull *pull)
+{
+  g_autoptr(GFile) root_dir = g_file_new_for_path (pull->root_dir_name);
+  g_autoptr(GFile) user_dir = g_file_new_for_path (pull->user_dir_name);
+
+  g_cancellable_cancel (pull->cancellable);
```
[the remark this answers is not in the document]
uajain: >Is there a chance it could be freed in unsuccessful_mount_timeout and then again in handle_deploy?

No, it can't happen. `unsuccessful_mount_timeout` will return the error and the pull will fallback to a `/var/tmp/..` location; and in the `handle_deploy`, it gets the fallback path and fails `ongoing_pull != NULL` condition. 

>I think there's a race because system_helper_ongoing_pull_free frees the members of the struct while the pull is still in the ongoing_pulls array, where another thread might find it and use or free it.

Still having trouble what scenario might run into this. Maybe something like this is better?
```
system_helper_ongoing_pull_free ( pull )
{
  gboolean removed = FALSE;
  G_LOCK (ongoing_pulls);
  removed = g_ptr_remove_array (ongoing_pulls, pull);
  G_UNLOCK (ongoing_pulls);

  if (removed)
  /* Free other members */
}
```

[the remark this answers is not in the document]
wjt: That check is too late: if two threads are operating on the same pointer, and one calls `free` before the other, then the other will be accessing freed memory.

To see a case where two threads may be operating on the same pointer, consider what would happen if a client calls `Deploy` and `CancelPull` for the same path at the same time.

 I think the following changes make sense:

* Split the bits of `_free` which are about tearing down state rather than freeing stuff into a separate `_teardown` or `_stop` function
* Add a refcount to the struct
* Take refs in the cases where the struct is passed as user_data to an async function
* Explicitly call `_teardown` on error paths, and also call it from `_unref` when we reach `0` refs and free the contents
* Make `get_ongoing_pull_by_dir` remove the pull from the array before returning it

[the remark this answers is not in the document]
uajain: > Make get_ongoing_pull_by_dir remove the pull from the array before returning it

Won't this break the criteria of idle exit ? idle timeout might see `ongoing_pulls` as none, and exits the system helper?

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
mwleeds: Yes, but probably worth putting that into a comment

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
mwleeds: I think it's safe to assume this was a side effect of the "double free FlatpakSystemHelperOngoingPull" problem

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
mwleeds: Yeah, according to a StackOverflow answer by the Fedora Project Leader, it's fine: https://unix.stackexchange.com/questions/71032/file-owner-must-belong-to-file-group

system-helper/flatpak-system-helper.c
```diff
@@ -172,6 +526,39 @@ handle_deploy (FlatpakSystemHelper   *object,
...
+      if (umount (ongoing_pull->user_dir_name) == -1)
+        {
+          glnx_throw_errno_prefix (&error, "Unmounting %s failed", ongoing_pull->user_dir_name);
+          g_dbus_method_invocation_return_gerror (invocation, error);
+          return TRUE;
+        }
+
+      if (ongoing_pull->bindfs)
+        {
+          g_subprocess_wait (ongoing_pull->bindfs, NULL, NULL);
```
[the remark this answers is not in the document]
uajain: Probably not. We already have `g_subprocess_wait_check_async` in place to return any errors.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +311,188 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+        {
+          g_dbus_method_invocation_return_error (invocation, G_DBUS_ERROR, G_DBUS_ERROR_LIMITS_EXCEEDED,
+                                                 "Maximum number of pulls per user exceeded for UID: %u", uid);
+          return TRUE;
+        }
+
+      // TODO: If there is a already partial pull available, return that.
+
+      flatpak_dir = g_file_get_path (flatpak_dir_get_path (system));
+      root_dir = g_mkdtemp_full (g_build_filename (flatpak_dir, "repo", "tmp", "flatpak-cache-XXXXXX", NULL), 0755);
```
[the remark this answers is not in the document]
uajain: You are right about the first question. Same filesystem etc.

It will get cleaned up `min-free-space-*` gating conditions are hit. They will clear out everything in `repo/tmp`. If they are not hit, they will / can be re-used to resume partial pulls. Checkout `check_partial_pull_availability` function.

system-helper/flatpak-system-helper.c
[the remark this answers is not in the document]
uajain: > Split the bits of _free which are about tearing down state rather than freeing stuff into a separate _teardown or _stop function
> Add a refcount to the struct
> Take refs in the cases where the struct is passed as user_data to an async function
> Explicitly call _teardown on error paths, and also call it from _unref when we reach 0 refs and free the contents

Addressed these.

> Make get_ongoing_pull_by_dir remove the pull from the array before returning it

Didn't address this (yet). I constantly find myself out of brain power to figure out any other approach here. I will probably need to take a fresh look after sometime.

system-helper/flatpak-system-helper.c
```diff
@@ -141,6 +371,343 @@ no_progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
...
+          root_dir = g_path_get_dirname (out_location);
+          user_dir = g_strdup_printf ("%s-bindfs", root_dir);
+          user_dir_file = g_file_new_for_path (user_dir);
+          if (!g_file_query_exists (user_dir_file, NULL))
+            g_mkdir_with_parents (user_dir, 0755);
+          child_repo_path = g_steal_pointer (&out_location);
+        }
+      else
+        {
+          g_autoptr(OstreeRepo) child_repo = NULL;
```
uajain: This right here, spits out a glib-critical. For some reason, this is having problems while `g_autoptr`-ized cleanup. I wonder if the problem is around `flatpak_dir_create_child_repo` below (given that I have changed function to a private API in this same commit)

bt:
```
F: GetChildRepoForPull 
F: Checking partial pull availability for ref org.gnome.Platform (commit: f6b4bc140bb2e2b8f72d1d2ae2109e4f951dfabfbe59102e22ba200ed2edeadf)

(/usr/libexec/flatpak-system-helper:2378): GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

Thread 5 "pool" received signal SIGTRAP, Trace/breakpoint trap.
[Switching to Thread 0x7fffe4812700 (LWP 2453)]
0x00007ffff538d4c1 in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
(gdb) bt
#0  0x00007ffff538d4c1 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#1  0x00007ffff538e7bd in g_logv () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007ffff538e91f in g_log () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#3  0x00007ffff6e3ec0f in  () at /lib/x86_64-linux-gnu/libostree-1.so.1
#4  0x00007ffff5669f8a in g_object_unref ()
    at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#5  0x0000555555570938 in glib_autoptr_cleanup_OstreeRepo (_ptr=<synthetic pointer>) at /usr/include/ostree-1/ostree-autocleanups.h:55
#6  0x0000555555570938 in handle_get_child_repo_for_pull (object=0x555555808760, invocation=<optimized out>, ref=0x7fffdc001bb0 "runtime/org.gnome.Platform/x86_64/3.30", commit=0x7fffdc001b60 "f6b4bc140bb2e2b8f72d1d2ae2109e4f951dfabfbe59102e22ba200ed2edeadf", arg_installation=<optimized out>)
    at system-helper/flatpak-system-helper.c:630
#7  0x00007fffef5ce038 in ffi_call_unix64 ()
    at /lib/x86_64-linux-gnu/libffi.so.6
#8  0x00007fffef5cda9a in ffi_call () at /lib/x86_64-linux-gnu/libffi.so.6
#9  0x00007ffff566575e in g_cclosure_marshal_generic ()
    at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#10 0x00007ffff5664f25 in g_closure_invoke ()
    at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#11 0x00007ffff56776e2 in  () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#12 0x00007ffff567f311 in g_signal_emitv ()
    at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
---Type <return> to continue, or q <return> to quit---
#13 0x00005555555b6064 in _flatpak_system_helper_skeleton_handle_method_call (connection=<optimized out>, sender=<optimized out>, object_path=<optimized out>, interface_name=0x7fffd80011b0 "org.freedesktop.Flatpak.SystemHelper", method_name=0x7fffd8005d70 "GetChildRepoForPull", parameters=<optimized out>, invocation=0x55555581a910, user_data=0x555555808760) at common/flatpak-dbus-generated.c:6780
#14 0x00007ffff597cb7e in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#15 0x00007ffff59337fd in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#16 0x00007ffff53b076e in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#17 0x00007ffff53afd75 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#18 0x00007ffff5126494 in start_thread (arg=0x7fffe4812700)
    at pthread_create.c:333
#19 0x00007ffff4e68acf in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:97
(gdb) 
```
@wjt @pwithnall  Any idea?

wjt: Nothing jumps out, sorry. Check out the [Debugging reference count problems](https://developer.gnome.org/gobject/stable/tools-refdb.html) tips from the GObject docs.

pwithnall: Nothing jumps out at me either. Try running under `valgrind` and see if it prints any other messages about memory corruption or reading/writing unallocated heap memory. The real problem might be caused by heap corruption elsewhere, and this is only a symptom of it. Make sure to use the [`glib.supp` suppressions file from GLib](https://gitlab.gnome.org/GNOME/glib/blob/master/glib.supp) with `--suppressions=glib.supp` to ignore known false positives.

uajain: I got it.

I didn't use the `ostree_repo_get_path` API properly. It has `[transfer none]` and I was assigning:
```
g_autoptr(GFile) repo_file = ostree_repo_get_path (child_repo);
```
Fixing it, fixes the critical too.
the `child_repo` is g_autoptr-ized too and hence `repo_file` pointer was being double-freed.

Thanks!
I will push fixup!

uajain:
This work got merged upstream in https://github.com/flatpak/flatpak/pull/2657

PR #123, opened 2018-10-01
Update symbols file

uajain:
Add flatpak_installation_get_min_free_space_bytes to symbols file

https://phabricator.endlessm.com/T23694

### DIFF
--- a/debian/libflatpak0.symbols
+++ b/debian/libflatpak0.symbols
@@ -27,6 +27,7 @@ libflatpak.so.0 libflatpak0 #MINVER#
  flatpak_installation_get_id@Base 0.8.0
  flatpak_installation_get_installed_ref@Base 0.5.2
  flatpak_installation_get_is_user@Base 0.5.2
+ flatpak_installation_get_min_free_space_bytes@Base 1.0.0-0endless1
  flatpak_installation_get_path@Base 0.5.2
  flatpak_installation_get_priority@Base 0.8.0
  flatpak_installation_get_remote_by_name@Base 0.5.2


debian/libflatpak0.symbols
```diff
@@ -27,6 +27,7 @@ libflatpak.so.0 libflatpak0 #MINVER#
  flatpak_installation_get_id@Base 0.8.0
  flatpak_installation_get_installed_ref@Base 0.5.2
  flatpak_installation_get_is_user@Base 0.5.2
+ flatpak_installation_get_min_free_space_bytes@Base 1.0.3
```
dbnicholson: Since this is a backport to our downstream version, I think you want to follow the version in the changelog, which is `1.0.0-0endless1`. Since we don't actually provide version 1.0.3 yet, any package that uses this symbol would get a dependency on a libflatpak0 package that doesn't exist.

uajain: Ok, so if we rebase(in future), do we need to change this back to `1.0.3`(?) because that's when the symbol will appear (as per upstream).

dbnicholson: If this really gets added as part of `1.0.3`, then debian will add it to their symbol file and we'll just drop our patch.

If we were being pedantic, we'd keep the version you're adding now since it will remain accurate in the future - the `flatpak_installation_get_min_free_space_bytes` symbol was added in our version `1.0.0-0endless1`, so any package that makes use of that symbol only needs to depend on a version of libflatpak0 greater than or equal to `1.0.0-0endless1`. That's the purpose of these files - they allow other packages to have fine-grained versioned dependencies based on what symbols they use.

That said, I'm sure we'll just drop our downstream patch and let the version be `1.0.3`. In our world, the cost of maintaining the additional patch is higher than the cost of having another package gain a higher versioned dependency.

PR #124, opened 2018-10-01
T23819: add API to allow us to squash fsyncs

ramcq:
Add _NO_TRIGGER to install/update/uninstall operations, and use utime() to update .changed file rather than creating a new one. Unmodified cherry-pick from upstream master.

https://github.com/flatpak/flatpak/pull/2161
https://phabricator.endlessm.com/T23856


common/flatpak-installation.c
```diff
@@ -1752,7 +1752,8 @@ flatpak_installation_install_full (FlatpakInstallation    *self,
                             ostree_progress, cancellable, error))
     goto out;
 
-  if (g_str_has_prefix (ref, "app"))
+  if ((flags & FLATPAK_INSTALL_FLAGS_NO_TRIGGERS) != 0 &&
```
pwithnall: The logic looks the wrong way round to me here. It checks whether `NO_TRIGGERS` is set and, if so, runs the triggers.

common/flatpak-installation.c
```diff
@@ -2663,3 +2663,30 @@ flatpak_installation_prune_local_repo (FlatpakInstallation *self,
 
   return flatpak_dir_prune (dir, cancellable, error);
 }
+
+/**
+ * flatpak_installation_run_triggers
```
wjt: Missing a colon after the function name here: I don't know if gtk-doc cares.

common/flatpak-installation.c
```diff
@@ -2663,3 +2663,30 @@ flatpak_installation_prune_local_repo (FlatpakInstallation *self,
...
+ * @self: a #FlatpakInstallation
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Run the trigger commands to update the files exported by the apps in
+ * @self. Should be used after one or more app install, upgrade or
+ * uninstall operations with the NO_TRIGGERS flag set.
```
wjt: Might be nice to link to the exact flag names, and/or link to this function from the docs for each `NO_TRIGGERS` flag

common/flatpak-installation.c
ramcq: 😂

common/flatpak-dir.c
```diff
@@ -2458,10 +2459,19 @@ flatpak_dir_mark_changed (FlatpakDir *self,
                           GError    **error)
 {
   g_autoptr(GFile) changed_file = NULL;
+  g_autofree char * changed_path = NULL;
 
   changed_file = flatpak_dir_get_changed_path (self);
+  changed_path = g_file_get_path (changed_file);
```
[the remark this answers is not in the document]
ramcq: Hmm, good suggestion, but looking at the docs, as g_file_peek_path is a cache, it would only confer a benefit if anything else used it - however this GFile is newly-allocated every time the changed path is accessed.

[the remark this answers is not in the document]
pwithnall: The docs also (correctly) say “It also avoids an extra duplicated string where possible”, which precisely true in the case of a local file.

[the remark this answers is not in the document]
pwithnall: i.e. In the case of a local file, no caching is used, and the internal path is returned directly.

common/flatpak-installation.c
pwithnall: gtk-doc does care.

common/flatpak-dir.c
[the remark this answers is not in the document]
ramcq: These ones don't: https://developer.gnome.org/gio/stable/GFile.html#g-file-peek-path

Anyway: we don't have that function.

common/flatpak-installation.c
ramcq: (This was a copy paste from @smspillaz preceding function! Which I have now also fixed.)

common/flatpak-dir.c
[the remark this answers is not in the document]
pwithnall: We should add that function (at some point). It’s pretty chill.

common/flatpak-installation.c
```diff
@@ -1752,7 +1752,8 @@ flatpak_installation_install_full (FlatpakInstallation    *self,
                             ostree_progress, cancellable, error))
     goto out;
 
-  if (g_str_has_prefix (ref, "app"))
+  if ((flags & FLATPAK_INSTALL_FLAGS_NO_TRIGGERS) == 0 &&
```
pwithnall: Nitpick: `!(flags & FLATPAK_INSTALL_FLAGS_NO_TRIGGERS)` would be a bit more readable.

common/flatpak-installation.c
```diff
@@ -2660,3 +2665,32 @@ flatpak_installation_prune_local_repo (FlatpakInstallation *self,
...
+/**
+ * flatpak_installation_run_triggers:
+ * @self: a #FlatpakInstallation
+ * @cancellable: (nullable): a #GCancellable
+ * @error: return location for a #GError
+ *
+ * Run the trigger commands to update the files exported by the apps in
+ * @self. Should be used after one or more app install, upgrade or
+ * uninstall operations with the #FLATPAK_INSTALL_FLAGS_NO_TRIGGERS,
+ * #FLATPAK_UPDATE_FLAGS_NO_TRIGGERS or #FLATPAK_UNINSTALL_FLAGS_NO_TRIGGERS
```
pwithnall: Nitpick: IIRC, gtk-doc wants enum members to be prefixed by `%`. `#` is for type names.

PR #125, opened 2018-10-02
create-usb: don't generate summary file

treitter:
Generating the summary requires acquiring an exclusive lock on the
OSTree repository which risks this process failing if another process,
like `ostree commit` or `ostree prune`, holds an exclusive lock for
longer than OSTree's core.lock-timeout-secs config option (which
defaults to only 30 seconds which is not always sufficient).

In testing directly and with Gnome Software calling this script to copy
the OS and dependencies to USB, this step seems to be unnecessary, so
we're best off skipping it.

https://phabricator.endlessm.com/T23422

### DIFF
--- a/app/flatpak-builtins-create-usb.c
+++ b/app/flatpak-builtins-create-usb.c
@@ -356,13 +356,6 @@ ostree_create_usb (GOptionContext *context,
     glnx_console_unlock (&console);
   }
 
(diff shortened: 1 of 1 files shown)


treitter:
Actually, we should merge this upstream first then cherry-pick.

I've opened https://github.com/flatpak/flatpak/pull/2165

mwleeds:
As I said upstream, this is not the right patch.

PR #128, opened 2018-10-12
flatpak-run: Add parental controls support for filtering apps

pwithnall:
Prevent the user from running a flatpak app if that app is filtered by
the parental controls applied to the user.

https://phabricator.endlessm.com/T24020


pwithnall:
This depends on eos-parental-controls with https://github.com/endlessm/eos-parental-controls/pull/7 applied.

I’ve just realised I need to do a corresponding `debian/` change. That will follow shortly.

pwithnall:
Debian changes in https://github.com/endlessm/flatpak/pull/129.

pwithnall:
bot, retest this please

pwithnall:
(Tests are currently failing because https://github.com/endlessm/eos-parental-controls/pull/7 has not been merged yet.)

common/flatpak-run.c
```diff
@@ -2853,13 +2865,31 @@ flatpak_run_app (const char     *app_ref,
...
+  g_autoptr(GAsyncResult) app_filter_result = NULL;
 
   struct stat s;
 
   app_ref_parts = flatpak_decompose_ref (app_ref, error);
   if (app_ref_parts == NULL)
     return FALSE;
 
+  /* Check that this user is actually allowed to run this app. */
+  epc_get_app_filter_async (NULL, getuid (), TRUE, cancellable,
```
wjt: Would it be so bad to add synchronous variants of these functions to the library? Every call-site I've seen so far (here and in `eos-parental-controls-client.py`) actually wants the function to be synchronous.

pwithnall: I was thinking that too, but was rushing to get this finished. I think it would make a good follow-up.

pwithnall: Thinking about it some more, it would add a significant amount of code to libeos-parental-controls, which couldn’t really be shared with the async implementation (since the library is essentially a chain of async calls, with no significant processing inbetween). I’m not sure that’s a maintenance burden worth taking on for the sake of avoiding some main context iterations in a client, given that I would expect the other clients to query/update the app filter synchronously.

I could be convinced otherwise, though, and would be happy to reassess this if the other client implementations turn out to need sync methods to call too.

pwithnall:
bot, retest this please

common/flatpak-run.c
wjt: One way to share code would be to implement the async version in terms of the sync one, in a GTask thread.

Not a merge blocker, I don't think. Just a thought.

PR #133, opened 2018-11-01
T24302 Ignore parental controls for nonexistent users

pwithnall:
https://phabricator.endlessm.com/T24302

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2807,17 +2807,6 @@ regenerate_ld_cache (GPtrArray    *base_argv_array,
   return glnx_steal_fd (&ld_so_fd);
 }
 
-static void
-async_result_cb (GObject      *obj,
-                 GAsyncResult *result,
-                 gpointer      user_data)
-{
-  GAsyncResult **result_out = user_data;
-
-  g_assert (*result_out == NULL);
-  *result_out = g_object_ref (result);
-}
-
 gboolean
 flatpak_run_app (const char     *app_ref,
                  FlatpakDeploy  *app_deploy,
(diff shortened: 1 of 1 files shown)


common/flatpak-run.c
```diff
@@ -2863,12 +2863,19 @@ flatpak_run_app (const char     *app_ref,
...
 
-  /* Check that this user is actually allowed to run this app. */
-  app_filter = epc_get_app_filter (NULL, getuid (), TRUE, cancellable, error);
-  if (app_filter == NULL)
-    return FALSE;
+  /* Check that this user is actually allowed to run this app. When running
+   * from the gnome-initial-setup session, an app filter might not be available. */
+  app_filter = epc_get_app_filter (NULL, getuid (), TRUE, cancellable, &my_error);
+  if (my_error != NULL &&
+      !g_error_matches (my_error, EPC_APP_FILTER_ERROR, EPC_APP_FILTER_ERROR_INVALID_USER))
```
GeorgesStavracas: Does that error really only represent "system users"?

pwithnall: In practice, yes, I think so. It’s basically raised if AccountsService doesn’t expose a D-Bus object for a user, which [happens for various users](https://gitlab.freedesktop.org/accountsservice/accountsservice/blob/204a4abe68eefac52ca05dda2096db0e0fde0bb7/src/user-classify.c#L147).

Actually, looking at that code has made me realise that we should probably be calling `org.freedesktop.Accounts.FindUserById` rather than building an object path dynamically. That would cause accountsservice to create an object for us for the gnome-initial-setup user, which would avoid this trouble.

In any case, I think the changes here are correct. flatpak should fail open if accountsservice can’t return data for the current user. The user should not be able to compromise accountsservice to force that kind of failure without having admin privileges.

PR #134, opened 2018-11-08
T24024 Check OARS parental controls for `flatpak install`

pwithnall:
https://phabricator.endlessm.com/T24024


pwithnall:
(The CI will fail for this until https://github.com/endlessm/eos-parental-controls/pull/13 is merged and built in OBS.)

pwithnall:
Fixed. Pay particular attention to how we get an `AsApp` from a flatpak ref in `get_appstream_for_ref()`. It uses the [appstream `<bundle/>` tag](https://www.freedesktop.org/software/appstream/docs/chap-CollectionData.html#tag-ct-bundle), which I think is OK — but there might be subtleties which I don’t properly understand.

common/flatpak-dir.c
```diff
@@ -7517,11 +7663,47 @@ flatpak_dir_install (FlatpakDir          *self,
...
+      app_filter = epc_get_app_filter (NULL, getuid (), TRUE, cancellable, error);
+      if (app_filter == NULL)
+        return FALSE;
+
+      AsContentRating *rating = appstream_get_latest_content_rating (app);
+
+      if (!appstream_check_rating (rating, app_filter))
+        return flatpak_fail (error,
+                             /* Translators: The placeholder is for an app ref. */
+                             _("%s has too extreme content for the current user"),
```
wjt: > %s has content too extreme for the current user

Or is that me hyper-correcting?

common/flatpak-dir.c
```diff
@@ -7502,6 +7505,149 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+    appstream_path = g_build_filename (install_path, "appstream", state->remote_name,
+                                       arch, "appstream.xml.gz",
+                                       NULL);
+  else
+    appstream_path = g_build_filename (install_path, "appstream", state->remote_name,
+                                       arch, "active", "appstream.xml.gz",
+                                       NULL);
+
+  g_autoptr(GFile) appstream_file = g_file_new_for_path (appstream_path);
+  as_store_from_file (store, appstream_file, NULL, cancellable, &local_error);
```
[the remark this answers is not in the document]
pwithnall: I have no idea; I can’t find much information or code about it. Based on https://blogs.gnome.org/hughsie/2018/09/20/speeding-up-appstream-mmaping-xml-using-libxmlb/, it looks like gnome-software has been ported (see the `xb_` calls in some of its plugins), and the binary XML library is an alternative to libappstream-glib, rather than a modification of it. In any case, this is not something we need to worry about until flatpak is ported to it; and if we share the flatpak code for loading appstream data (as per your comment below), we should get it for free (unless we need to port away from libappstream-glib). Sigh.

common/flatpak-dir.c
```diff
@@ -7502,6 +7505,149 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+
+  if (arch == NULL)
+    arch = flatpak_get_arch ();
+
+  g_autofree char *install_path = g_file_get_path (flatpak_dir_get_path (self));
+  g_autoptr(AsStore) store = as_store_new ();
+  g_autofree char *appstream_path = NULL;
+  g_autofree char *unique_id = NULL;
+
+  if (flatpak_dir_get_remote_oci (self, state->remote_name))
```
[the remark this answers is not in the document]
pwithnall: Seems reasonable. I’ll take a look at that.

common/flatpak-dir.c
```diff
@@ -7517,11 +7663,47 @@ flatpak_dir_install (FlatpakDir          *self,
...
+      g_autoptr(EpcAppFilter) app_filter = NULL;
+      app_filter = epc_get_app_filter (NULL, getuid (), TRUE, cancellable, error);
+      if (app_filter == NULL)
+        return FALSE;
+
+      AsContentRating *rating = appstream_get_latest_content_rating (app);
+
+      if (!appstream_check_rating (rating, app_filter))
+        return flatpak_fail (error,
+                             /* Translators: The placeholder is for an app ref. */
```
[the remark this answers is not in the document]
pwithnall: Short of pointing to the documentation or copying some of the documentation into the translation comment, I’m not sure how to improve the comment. An ‘app ref’ is a pretty fundamental bit of flatpak, and would pop up in a lot of translatable strings if any of them had translation comments.

Yeah, `G_IO_ERROR_FAILED` is sadly very pervasive in flatpak. Fighting it seems to achieve nothing.

common/flatpak-dir.c
pwithnall: Meh, I’ll change it.

common/flatpak-dir.c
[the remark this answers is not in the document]
pwithnall: Update pushed which does that, and upstream PR opened here: https://github.com/flatpak/flatpak/pull/2296

common/flatpak-dir.c
```diff
@@ -7502,6 +7505,179 @@ flatpak_dir_create_system_child_repo (FlatpakDir   *self,
...
+                                         cancellable, error))
+    return NULL;
+
+  unique_id = as_utils_unique_id_build (AS_APP_SCOPE_UNKNOWN,
+                                        AS_BUNDLE_KIND_FLATPAK,
+                                        NULL,  /* origin */
+                                        AS_APP_KIND_DESKTOP,
+                                        flatpak_ref_get_name (ref),
+                                        flatpak_ref_get_branch (ref));
+  AsApp *app = as_store_get_app_by_unique_id (store, unique_id,
```
wjt: If the same app is available for 2 different architectures, what gets returned here? For instance, you can install and run i386 flatpaks on an x86_64 system. Can a user circumvent parental controls by installing the i386 one?

pwithnall: I believe flatpak generates the `appstream.xml` file for the repository from the app’s OSTree commit, and it injects the `<bundle>` elements into it. Therefore the appstream is always generated separately for each architecture of an app, and so should always contain the right OARS data. See [`extract_appstream()` in `flatpak-utils.c`](https://github.com/flatpak/flatpak/blob/bf7af547aa65eeee0f9c7a4a73e2c88eb4b84693/common/flatpak-utils.c#L3390).

wjt: Is the arch name included in the blacklist entry?

pwithnall: Blacklist entry? The `appstream.xml` file is stored per-arch in `/var/lib/flatpak/appstream/${remote_name}/${arch}/appstream.xml{,.gz}`.

wjt: My concern was that you could circumvent the filter by installing an app from a foreign arch (i386 on x86_64) but I now see that the appstream DB that gets loaded is based on the arch being run.

PR #135, opened 2018-11-13
dir: Allow app filter to disallow app installation entirely

pwithnall:
This is independent from the app-install polkit action, which determines
whether an unprivileged user may install an app system-wide. Being
stored in accountsservice, the new boolean is also easier to set
per-user without having to programmatically write a polkit JS policy
file which handles multiple users (and parse it back again).

It overrides the OARS values.

https://phabricator.endlessm.com/T24457

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7726,6 +7726,8 @@ flatpak_dir_install (FlatpakDir          *self,
 
       AsContentRating *rating = appstream_get_latest_content_rating (app);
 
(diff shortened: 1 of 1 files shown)


GeorgesStavracas:
retest this please

pwithnall:
(It won’t build successfully until the eos-parental-controls with the new API is built in OBS, which should be done now.)

GeorgesStavracas:
It did work

PR #137, opened 2018-12-05
debian: Bump eos-parental-controls-0 dependency for new symbols

pwithnall:
We need epc_app_filter_is_user_installation_allowed().

https://phabricator.endlessm.com/T24457

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -28,7 +28,7 @@ Build-Depends:
  libcap-dev,
  libdw-dev,
  libelf-dev,
- libeos-parental-controls-0-dev,
+ libeos-parental-controls-0-dev (>= 0+dev53.de16300bem1),
  libgirepository1.0-dev,
  libglib2.0-dev (>= 2.44),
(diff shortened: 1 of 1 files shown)


debian/control
```diff
@@ -28,7 +28,7 @@ Build-Depends:
  libcap-dev,
  libdw-dev,
  libelf-dev,
- libeos-parental-controls-0-dev,
+ libeos-parental-controls-0-dev (>= 0+dev53.de16300bem1),
```
wjt: Why is this needed? Won't the newest version always be included in the Jenkins and OBS build environments, and in the image?

pwithnall: Thinking about the recent runtime failures from various people who’ve built the Shell locally and haven’t updated their build dependencies, I thought it might be wise to try and avoid that here too. If you strongly feel it isn’t necessary then I’m happy to drop the PR>

wjt: I suppose it can't hurt!

PR #142, opened 2019-01-22
T25088 common: Update AppStream data before doing an installation

pwithnall:
It’s needed in order to make a correct and informed decision about
parental controls filters.

https://phabricator.endlessm.com/T25088


pwithnall:
CR ready again, now rebased on our latest flatpak.

common/flatpak-installation.c
```diff
@@ -1903,6 +1903,11 @@ flatpak_installation_install_full (FlatpakInstallation    *self,
   else
     ostree_progress = ostree_async_progress_new_and_connect (no_progress_cb, NULL);
 
+  /* AppStream data is needed for parental controls decisions. */
+  if (!flatpak_dir_update_appstream (dir_clone, remote_name, arch, NULL,
```
mwleeds: I'm not sure if this throws a wrench into offline USB app installs (or even offline app installs using the `FLATPAK_INSTALL_FLAGS_NO_DEPLOY` and `FLATPAK_INSTALL_FLAGS_NO_PULL` flags) so I added an AC for that to the ticket.

mwleeds: Another reason that a failure of `flatpak_dir_update_appstream` maybe should be treated as nonfatal is that some remotes don't provide appstream data. See https://github.com/flatpak/flatpak/issues/1615 and https://github.com/flatpak/flatpak/issues/427

pwithnall: Indeed. So is there anything to do here? As I read them, all your review comments cancel each other out.

pwithnall: I don’t think this PR will cause any problems with USB updates which weren’t already there. The parental controls stuff will cause an app installation to fail if the appstream data is not available to indicate the OARS rating of that app. That hasn’t changed with this PR.

I suspect there will be (and are already) problems if you try to install an app from a USB stick with a collection ID for which you don’t already have cached appstream data coming from an online remote. If the USB stick provides the appstream branch, things should work properly. The `create-usb` flatpak builtin does already add the `appstream` and `appstream2` branches, so things should work properly end-to-end.

mwleeds: I don't see how they cancel each other out. If appstream data is not available from every remote, shouldn't a failure of `flatpak_dir_update_appstream()` in `flatpak_installation_install_full()` not be fatal to the installation?

pwithnall:
Closing unmerged as per [this comment](https://phabricator.endlessm.com/T25088#740957).

PR #145, opened 2019-01-29
Revert "Stop linking libflatpak against listappstream-glib"

mwleeds:
This reverts commit f9c6a769ef30a3a33038ec91f080197a0bf418a5.

This is necessary because we reverted the upstream commit 9dff4bbb8 in
order to support the parental controls feature.

https://phabricator.endlessm.com/T25194


treitter:
Looks good.

PR #143, opened 2019-01-28
Backports from 1.2.0

mwleeds:
These are the upstream commits that look important, of the ones we didn't get in the 1.1.3 rebase.

https://phabricator.endlessm.com/T25238

PR #146, opened 2019-01-29
Fix compiler warnings/errors and unit tests

mwleeds:
https://phabricator.endlessm.com/T25194

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -2322,16 +2322,16 @@ add_appdata_to_deploy_data (GVariantBuilder *metadata_builder,
 }
 
 static GVariant *
-flatpak_dir_new_deploy_data (FlatpakDir *self,
-                             GFile      *deploy_dir,
-                             GVariant   *commit_metadata,
-                             GKeyFile   *metadata,
-                             const char *id,
-                             const char *origin,
-                             const char *commit,
-                             char      **subpaths,
-                             guint64     installed_size,
-                             char      **previous_ids)
+flatpak_dir_new_deploy_data (FlatpakDir         *self,
+                             GFile              *deploy_dir,
+                             GVariant           *commit_metadata,
+                             GKeyFile           *metadata,
+                             const char         *id,
+                             const char         *origin,
+                             const char         *commit,
+                             char              **subpaths,
+                             guint64             installed_size,
+                             const char * const *previous_ids)
 {
   char *empty_subpaths[] = {NULL};
   GVariantBuilder metadata_builder;
@@ -2366,7 +2366,7 @@ flatpak_dir_new_deploy_data (FlatpakDir *self,
 
   if (previous_ids)
     g_variant_builder_add (&metadata_builder, "{s@v}", "previous-ids",
-                           g_variant_new_variant (g_variant_new_strv ((const char * const *) previous_ids, -1)));
+                           g_variant_new_variant (g_variant_new_strv (previous_ids, -1)));
 
   add_appdata_to_deploy_data (&metadata_builder, deploy_dir, id);
 
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7959,10 +7959,31 @@ flatpak_dir_user_check_parental_controls (FlatpakDir    *self,
 {
   g_autoptr(AsApp) app = NULL;
   g_autoptr(GError) local_error = NULL;
+  const char *on_session = g_getenv ("FLATPAK_SYSTEM_HELPER_ON_SESSION");
+  g_autoptr(GDBusConnection) dbus_connection = NULL;
 
   /* The system helper implements its own checks via polkit. */
   if (!self->user)
     return TRUE;
+
+  dbus_connection = g_bus_get_sync (G_BUS_TYPE_SYSTEM, cancellable, &local_error);
+  if (dbus_connection == NULL)
+    {
+      if (on_session != NULL)
+        {
+          /* FIXME: Instead of skipping the parental controls check in the test
+           * environment, make a mock service for it.
+           * https://phabricator.endlessm.com/T25340 */
+          g_debug ("Skipping parental controls check for %s since the "
+                   "system bus is unavailable in the test environment", ref);
+          return TRUE;
+        }
+      else
+        {
+          g_propagate_error (error, g_steal_pointer (&local_error));
+          return FALSE;
+        }
+    }
 
   /* Get the AppStream data for this app, so we can check its content rating. */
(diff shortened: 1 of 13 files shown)


mwleeds:
This branch now includes the backports from #143 in addition to two compilation fixes

mwleeds:
bot retest this please

mwleeds:
This is now CR Ready and has a few different kinds of commits.

**Cherry-picks from upstream:**
```
3fa58740 build-commit-from: Fix handling of ostree.collection-refs-binding
2d5c67bd validate-icon: Search the path for bwrap
beaf3429 build-commit-from: Support --extra-collection-id
d5e5d3a2 run: Generate /run/host/font-dirs.xml config file
47e80d48 cli transaction: Use raw mode
f219c0b2 Add utilities to switch in and out of raw mode
42848fd8 icon-validator: Use readlink() properly
```
**Cherry-picks from Debian:** (being upstreamed in https://github.com/flatpak/flatpak/pull/2661)
```
67290033 make-test-*: Don't sandbox the icon validator
1df44284 build-export: Allow sandboxing on icon validator to be disabled
```
**Downstream fixes for compiler warnings/errors and unit tests:**
```
4c6cdba2 dir: Avoid parental controls check in testlibrary
0f8ea887 system-helper: Properly use dir_get_system()
945e29e2 Fix a compiler warning in use of flatpak_dir_new_deploy_data()
```

common/flatpak-dir.c
```diff
@@ -7959,11 +7959,29 @@ flatpak_dir_user_check_parental_controls (FlatpakDir    *self,
...
   if (!self->user)
     return TRUE;
 
+  dbus_connection = g_bus_get_sync (G_BUS_TYPE_SYSTEM, cancellable, &local_error);
+  if (dbus_connection == NULL)
+    {
+      if (on_session != NULL)
+        {
+          g_debug ("Skipping parental controls check for %s since the "
+                   "system bus is unavailable in the test environment", ref);
```
wjt: Doesn't this allow a user to bypass parental controls by running:

```
FLATPAK_SYSTEM_HELPER_ON_SESSION=1 DBUS_SYSTEM_BUS_ADDRESS=/dev/null flatpak install --user flathub very.violent.game
```

mwleeds: Ah, I thought about that possibility but didn't know about `DBUS_SYSTEM_BUS_ADDRESS`

mwleeds: The cheap way to work around that would be to check if `DBUS_SYSTEM_BUS_ADDRESS` is set when deciding whether to propagate the error, but there are probably legitimate reasons to set that

mwleeds: Maybe we can copy what's done in the tests in `eos-parental-controls` but it will require a dependency on `libglib-testing`

mwleeds: Upon further investigation I don't know of a reasonable way to do this. If `testlibrary` creates the `GDBusConnection` based on a `GtDBusQueue` there's no reasonable way for it to pass that to libflatpak where it's needed. And if libflatpak creates the `GDBusConnection` we would still need a way to distinguish the test environment from production, like a `FlatpakInstallFlags` value but that seems messy.

I wonder if we could just assume kids won't know to set these environment variables? Couldn't a determined user also just compile a version of flatpak that doesn't do parental controls checks?

common/flatpak-dir.c
```diff
@@ -7402,7 +7402,7 @@ flatpak_dir_deploy (FlatpakDir          *self,
                                              checksum,
                                              (char **) subpaths,
                                              installed_size,
-                                             previous_ids);
+                                             (char **)previous_ids);
```
pwithnall: Shouldn’t you just make the `previous_ids` argument in the definition of `flatpak_dir_new_deploy_data()` a `const gchar * const *`?

system-helper/flatpak-system-helper.c
```diff
@@ -1201,12 +1201,13 @@ handle_update_summary (FlatpakSystemHelper   *object,
...
  * a policy here. */
 static gboolean
-authorize_deploy_add_polkit_details (const gchar    *installation,
-                                     const gchar    *ref,
-                                     const gchar    *origin,
-                                     PolkitSubject  *subject,
-                                     PolkitDetails  *details,
-                                     GError        **error)
+authorize_deploy_add_polkit_details (const gchar           *installation,
+                                     GDBusMethodInvocation *invocation,
```
pwithnall: This “system-helper: Properly use dir_get_system()” commit should have something in its commit message saying that it needs to be squashed into “system-helper: Move parental controls checks to system-helper” on the next rebase.

common/flatpak-dir.c
pwithnall: You either have to provide a way to disable parental controls checks, or mock up the parental controls D-Bus service (which is actually part of `org.freedesktop.Accounts`). While I’ve got a suggestion for how to do the latter, it’s a few hours’ work so might make sense to leave until after the rebase. If that’s the case, then I can’t see anything immediately better than your current patch (plus a `FIXME` comment, plus a follow-up issue).

My longer-term suggestion:
 1. Drop `FLATPAK_SYSTEM_HELPER_ON_SESSION` and instead set `DBUS_SYSTEM_BUS_ADDRESS`, just to simplify the code a bit. You can get the address to set it to from `g_test_dbus_get_bus_address()` in `testlibrary.c`.
 2. Add a service file in `tests/services` (`services_dir` in `testlibrary.c`) which advertises `org.freedesktop.Accounts` and has its `Exec=` line as a simple Python program which mocks that service just enough to allow parental controls to pass. That would mean listing one or two dummy accounts, and returning the default parental controls properties for them. It would be about 100 lines, I’m guessing, and wouldn’t need to change its behaviour for different flatpak tests.

If we ever wanted to add unit tests for parental controls interaction between the system helper and accounts-daemon, that’s a bigger problem which is out of scope here. It might involve `GtDBusQueue`, although as the amount of code under test gets bigger, `GtDBusQueue` becomes less appropriate (as you end up setting up a huge amount of state for each test). `python-dbusmock` might be more appropriate, to implement a mock accounts-daemon service whose behaviour can be tweaked by other D-Bus methods it exposes.

pwithnall:
The rest of the backports look good to me,, and I’ve checked there’s no other commits upstream (as of right now) which look relevant.

common/flatpak-dir.c
mwleeds: yes you're right

common/flatpak-dir.c
mwleeds: Added the FIXME comment

pwithnall:
fixups LGTM.

PR #149, opened 2019-02-11
Backport security fix from Flatpak 1.0.7

wjt:
Upstream commit message:

> Don't expose /proc when running apply_extra
> 
> As shown by CVE-2019-5736, it is sometimes possible for the sandbox
> app to access outside files using /proc/self/exe. This is not
> typically an issue for flatpak as the sandbox runs as the user which
> has no permissions to e.g. modify the host files.
> 
> However, when installing apps using extra-data into the system repo
> we *do* actually run a sandbox as root. So, in this case we disable mounting
> /proc in the sandbox, which will neuter attacks like this.

Modified to fix a minor conflict in the FlatpakRunFlags hunk due to FLATPAK_RUN_FLAG_CANBUS not being present in our 1.0.x branch. This is the eos3.5 counterpart to https://github.com/endlessm/flatpak/pull/148.

https://phabricator.endlessm.com/T25408

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -2340,9 +2340,13 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
     "# Disable user pkcs11 config, because the host modules don't work in the runtime\n"
     "user-config: none\n";
 
+  if ((flags & FLATPAK_RUN_FLAG_NO_PROC) == 0)
+    flatpak_bwrap_add_args (bwrap,
+                            "--proc", "/proc",
+                            NULL);
+
   flatpak_bwrap_add_args (bwrap,
                           "--unshare-pid",
-                          "--proc", "/proc",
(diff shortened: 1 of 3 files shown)


common/flatpak-run.c
```diff
@@ -2340,9 +2340,13 @@ flatpak_run_setup_base_argv (FlatpakBwrap   *bwrap,
     "# Disable user pkcs11 config, because the host modules don't work in the runtime\n"
     "user-config: none\n";
 
+  if ((flags & FLATPAK_RUN_FLAG_NO_PROC) == 0)
+    flatpak_bwrap_add_args (bwrap,
+                            "--proc", "/proc",
```
wjt: I have verified with `flatpak run --command=ls org.vim.Vim /proc` that `/proc` shows up fully populated when running an app; and with `find /proc` in an `apply_extra` script that it is missing while applying extra-data.

PR #156, opened 2019-05-15
T26715: Add support for entitlements servers

dnohales:
So here is a draft of what I'm working on, basically, before pulling the the ref from the repo, it checks if the repo has an Entitlements server associated through the `xa.api-server-url` config option. If it does, it contacts the entitlements server to get the token using the following API call (that I mocked in my local `flatpak-entitlements-server`):

```
<api-server-url>/token?ref=app/org.gnome.eog/x86_64/stable&repo_url=http://localhost:8080/repo/stable
```

The server then returns a token valid for 1 hour. Then OSTree uses that token to retrieve the objects.

Obviously, this is a draft, the following is pending:

- [ ] Find a way to pass different pull options per remote for the `ostree_repo_find_remotes/ostree_repo_pull_from_remotes`, may require more changes in ostree.
- [x] Add `xa.api-server-url` everywhere else in Flatpak since these kind of options are scattered everywhere in the code.
- [ ] Add `xa.api-server-url` to docs.
- [x] Save tokens to cache.
- [x] Use the computer login credentials to do the request to the entitlements server.

https://phabricator.endlessm.com/T26715


dbnicholson:
I think rather than using a repo config option (which are tricky to set reliably besides the first time the remote is added), I think it should just try the request unauthenticated and then the repo server will send back a custom HTTP header with the 401 response.

dnohales:
Ok, so the repo will respond with a 401 response including in the payload or header (`WWW-Authenticate` maybe) the URI to request the token?

Then I suppose all the authorization workflow is going to be implemented ostree side, not Flatpak, right?

dbnicholson:
I forgot that you don't actually get the raw HTTP response back from a failed pull, so maybe that's not the way to go. Let's just continue for now with the implementation you have with the repo config option. That said, there's nothing actually flatpak specific about this feature. It's all about authorized access to ostree repo objects.

ramcq:
I don't think the name is great (I don't really like the term entitlement, it usually means somebody is acting like an asshole) - maybe just call it token-server, api-server, access-server, etc.

But I think an option is the right approach, it's going to be way quicker and more efficient if you just know up front that you need to use that for every commit/delta. Round trips harm throughput.

The other thought is we should also send the collection ID if configured - that takes precedence over the repo URL for uniquely identifying a ref.

dnohales:
I think I could get the response back, but that should be done from inside ostree.

dnohales:
@dbnicholson @ramcq Updated

dnohales:
Jenkins test this please

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+        {
+          g_prefix_error (error, "Unable to get API server access token: ");
+          return NULL;
+        }
+
+      token_url = g_strconcat (api_server_url,
+                               "/token?ref=",
+                               ref_to_fetch,
+                               "&repo_url=",
+                               remote_url,
```
pwithnall: These parameters should be URI-escaped before building a URI with them. `SoupURI` provides an API for building URIs.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+                               NULL);
+
+      token_url_checksum = g_compute_checksum_for_string (G_CHECKSUM_SHA1,
```
pwithnall: SHA-1 has been considered [increasingly insecure since 2005](https://en.wikipedia.org/wiki/SHA-1#Attacks), and hence its use here should be justified with a comment.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+                                                          strlen (token_url));
+
+      cache_dir = flatpak_ensure_system_user_cache_dir_location (error);
+      if (!cache_dir)
+        {
+          g_prefix_error (error, "Unable to get system cache dir: ");
+          return NULL;
+        }
+
+      token_cache_filename = g_strconcat (g_file_get_path (cache_dir),
```
pwithnall: Build filenames using `g_build_filename()`.

The `g_file_get_path()` call here leaks its return value. Use `g_file_peek_path()` if flatpak depends on a new-enough GLib; otherwise assign to a local variable and free the result.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+                                          "/api-server-token-",
+                                          token_url_checksum,
+                                          NULL);
+      token_cache_file = g_file_new_for_path (token_cache_filename);
+      token_cache_file_info = g_file_query_info (token_cache_file,
```
pwithnall: Need to handle failure from `g_file_query_info()`.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+                                                 G_FILE_ATTRIBUTE_TIME_MODIFIED,
+                                                 G_FILE_QUERY_INFO_NONE,
+                                                 cancellable, NULL);
+      g_file_info_get_modification_time (token_cache_file_info, &token_cache_mtime);
+      g_get_current_time (&current_time);
+
+      /* Tokens expires in one hour, we won't reuse tokens that are close
+       * to expire, that's why we check for at most 3400 seconds old tokens */
+      if (g_file_query_exists (token_cache_file, cancellable) &&
```
pwithnall: No need to call `g_file_query_exists()`; you can do that as part of the `g_file_query_info()` call above.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+        {
+          if (!g_file_load_contents (token_cache_file,
+                                     cancellable,
+                                     &token_buffer,
+                                     NULL, NULL,
+                                     error))
+            {
+              g_warning ("Unable to read token cache file %s: %s",
+                         token_cache_filename,
+                         (*error)->message);
```
pwithnall: This will crash if the caller passes `NULL` to the `GError **error` argument of `get_repo_bearer_token()`. Use a `g_autoptr(GError) local_error` instead, and use `g_propagate_error (error, g_steal_pointer (&local_error))` to propagate that error to the caller when necessary.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+      while (request_read_all_result == NULL)
+        g_main_context_iteration (context, TRUE);
+
+      if (!g_input_stream_read_all_finish (request_stream,
+                                           request_read_all_result,
+                                           &token_size,
+                                           error))
+        {
+          g_prefix_error (error, "Unable to read token from server: ");
+          g_clear_pointer (&token_buffer, g_free);
```
pwithnall: Rather than freeing `token_buffer` on every failure code path, define it as `g_autofree gchar *token_buffer = NULL`, and use `return g_steal_pointer (&token_buffer)` on the success code paths where you want to return it to the caller without freeing.

Note that also you should use `guint8 *` as the type for `token_buffer` unless it’s guaranteed to be a UTF-8 string. Otherwise you have to assume it’s arbitrary binary data.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+                                                 error);
+      if (!request_stream)
+        {
+          g_prefix_error (error, "Unable to create repo token Soup request: ");
+          return NULL;
+        }
+
+      if (!SOUP_STATUS_IS_SUCCESSFUL (http_message->status_code))
+        {
+          *error = g_error_new (G_IO_ERROR, G_IO_ERROR_FAILED,
```
pwithnall: You can’t assume that `error` will be non-`NULL`. Use `g_set_error (error, …)` to set the error.

See the [summary documentation on using `GError`](https://developer.gnome.org/glib/stable/glib-Error-Reporting.html#gerror-rules).

common/flatpak-dir.c
pwithnall: It seems it’s just used here to conveniently create an escaped filename for the cache. What’s the reasoning for using this scheme for naming files in the cache?

Would it be better to explicitly build the filename from canonical versions of the parameters in the URI, excluding stuff like `/token?ref=` which could change in future (and unnecessarily invalidate clients’ cached tokens)? i.e. Build the path as `${cache_dir}/api-server-tokens/${escaped_api_server_url}/${remote_url}/${ref}`.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+        {
+            {  "serial", SECRET_SCHEMA_ATTRIBUTE_STRING },
+            {  "NULL", 0 },
+        }
+    };
+    return &schema;
+}
+
+static char *
+get_api_server_access_token (GCancellable  *cancellable,
```
pwithnall: As you say, the way to get the machine-specific credentials is pretty Hack-specific. I am not sure upstream would be happy with a mandatory libsecret dependency. You could just load the token from a well-known file in the user’s home directory, for now, and perhaps carry a patch to use libsecret downstream in Endless?

The design document isn’t particularly clear on the design for authenticating with the entitlement server, so I can’t really comment on what (if anything) would be more appropriate here.

common/flatpak-dir.c
```diff
@@ -4018,19 +4022,231 @@ repo_get_remote_collection_id (OstreeRepo *repo,
...
+      request = soup_session_request_http (dir->soup_session, "GET",
+                                           token_url, error);
+      if (!request)
+        {
+          g_prefix_error (error, "Unable to create repo token Soup request: ");
+          return NULL;
+        }
+
+      http_message = soup_request_http_get_message (request);
+      bearer_header = g_strconcat ("Bearer: ", access_token, NULL);
```
pwithnall: As I understand it, the [`Authorization` header doesn’t require an additional `:`](https://tools.ietf.org/html/rfc1945#section-11), so this should be `"Bearer "`.

This `access_token` is the same one which just came straight out of libsecret, so is presumably a fairly permanent secret which is shared between the device and the server. That seems like a bad idea: if anyone ever got a copy of the `access_token` for a particular machine, they could use it in perpetuity to access all the apps this user has access to. Instead, the code should send a fresh hash of the shared secret — something along the lines of `hmac(payload=current_date, key=access_token)`, but with a little more thought put into it than my quick sketch here.

Unless `access_token` is something set by some Hack-specific service on the machine, and rotated regularly? Basically I don’t think I know enough about the architecture here to comment for sure, but my ramblings above should outline what I think should be avoided.

common/flatpak-dir.c
```diff
@@ -4174,37 +4403,40 @@ repo_pull (OstreeRepo                           *self,
...
+                               dirs_to_pull, current_checksum, force_disable_deltas,
+                               flags, progress, cancellable);
       pull_options = g_variant_ref_sink (g_variant_builder_end (&pull_builder));
 
       context = flatpak_main_context_new_default ();
 
       if (results_to_fetch == NULL)
         {
-          ostree_repo_find_remotes_async (self, (const OstreeCollectionRef * const *) collection_refs_to_fetch,
+          ostree_repo_find_remotes_async (repo, (const OstreeCollectionRef * const *) collection_refs_to_fetch,
```
pwithnall: Note: There are other calls to `ostree_repo_find_remotes_async()` in flatpak which you’ll also need to modify to send the right tokens.

pwithnall: Regarding sending the right headers to the right servers, the issue is that OSTree’s P2P support is deliberately designed to allow requesting many refs from many servers in parallel, to try and reduce the number of network requests. I don’t think it’s possible to set a single header which would serve as a suitable token for all the requests, so I think the solution has to be for the calling code in flatpak to provide OSTree with all the tokens it could possibly need for a single `ostree_repo_find_remotes_async()` call, and information on when to use each of them.

Since each token is valid for a single ref, I suggest you could implement a `ref-http-headers` option (that’s just one suggested name; feel free to come up with a better one) in OSTree which is a map of ref-name to an `a(ss)` array of headers to send whenever requesting the commit object for that ref-name.

This would need to be proposed in upstream OSTree, need implementing there, and then it could be used in flatpak.

common/flatpak-dir.c
ramcq: It's refreshed by the hack account service and has its own lifecycle etc which is maintained for us. In place of the hack token it could just as easily be any other OAuth thing from GOA or similar.

pwithnall: That sounds appropriate then. Since it’s a (short life) shared secret, rather than a challenge and response, it can only be sent over HTTPS. So we should probably error out if the repository server returns a non-HTTPS URI for the token server.

common/flatpak-dir.c
dylanmccall: Would it make sense to store a server access token per API server? It feels awkward to me that we take an `api_server_url` in `get_repo_bearer_token` but we assume whatever API server we're talking to is going to use whatever token we're using.

Also +1 for using libsecret like this being a little too Hack-specific :) Perhaps we could split get_api_server_access_token out of flatpak-dir.c to make this a little easier to reason about, then upstream can have a naive implementation that uses a plain file per API server (probably the same scheme you use for cached tokens), while downstream can patch in the libsecret + serial number stuff for the Hack API server.

app/flatpak-builtins-remote-modify.c
```diff
@@ -167,6 +169,13 @@ get_config_from_opts (FlatpakDir *dir, const char *remote_name, gboolean *change
       *changed = TRUE;
     }
 
+  if (opt_api_server_url)
+    {
+      g_key_file_set_string (config, group, "xa.api-server-url", opt_api_server_url);
```
dylanmccall: I'm worried I'm late to this and missing some context, but the name api-server-url (here and elsewhere) seems overly generic. This is specifically the entitlements API, right? Could we call it entitlements-server-url?

pwithnall: Context/Bikeshed here: https://docs.google.com/a/endlessm.com/document/d/1zE_QbB6mtdhjH5bsFdf9kPrYvjKAMBRl4uSRYDt-BqQ/edit?disco=AAAAC8Pzh-s

mwleeds:
Closing since this was [implemented upstream in a different way](https://github.com/flatpak/flatpak/pull/3167).

PR #160, opened 2019-06-05
common: Do not convert G_IO_ERROR_CANCELLED errors to FLATPAK_ERROR d…

uajain:
…omain when fetching remote metadata.

While trying back and forth searching in gnome-software, G_IO_ERROR_CANCELLED
errors end up in the UI via the error-banner. It is quite a frustating UX.

GNOME Software already filters out G_IO_ERROR_CANCELLED errors from ending up
in the error-banner/UI; but still somehow many of these errors still were
ending up in the error-banner. This was due to the fact that these errors
were converted to FLATPAK_ERROR domain with code as FLATPAK_ERROR_INVALID_DATA
and hence were not filtered out by gnome-software.

Also, it's actually not right to have something like FLATPAK_ERROR_CANCELLED and
teach gnome-software to detect it. Hence, the approach here is to preserve the
error domain as G_IO_ERROR and error code as G_IO_ERROR_CANCELLED.

https://phabricator.endlessm.com/T23928

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -275,8 +275,18 @@ flatpak_remote_state_ensure_summary (FlatpakRemoteState *self,
                                      GError            **error)
 {
   if (self->summary == NULL)
-    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Unable to load summary from remote %s: %s"), self->remote_name,
-                               self->summary_fetch_error != NULL ? self->summary_fetch_error->message : "unknown error");
+    {
+      if (self->summary_fetch_error != NULL &&
+          g_error_matches (self->summary_fetch_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        {
+          g_propagate_prefixed_error (error, g_error_copy (self->summary_fetch_error),
+                                      _("Unable to load summary from remote %s: "), self->remote_name);
+          return FALSE;
+        }
+
+      return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Unable to load summary from remote %s: %s"), self->remote_name,
+                                 self->summary_fetch_error != NULL ? self->summary_fetch_error->message : "unknown error");
+    }
 
   return TRUE;
 }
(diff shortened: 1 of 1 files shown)


common/flatpak-dir.c
```diff
@@ -275,8 +275,18 @@ flatpak_remote_state_ensure_summary (FlatpakRemoteState *self,
                                      GError            **error)
 {
   if (self->summary == NULL)
-    return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Unable to load summary from remote %s: %s"), self->remote_name,
-                               self->summary_fetch_error != NULL ? self->summary_fetch_error->message : "unknown error");
+    {
+      if (self->summary_fetch_error != NULL &&
+          g_error_matches (self->summary_fetch_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+        {
+          g_propagate_prefixed_error (error, g_steal_pointer (&self->summary_fetch_error),
```
pwithnall: Should we steal the `summary_fetch_error` here? The other code path doesn’t steal it; it just reads the error. I think we might want to `g_error_copy (self->summary_fetch_error)` instead.

common/flatpak-dir.c
```diff
@@ -291,9 +301,30 @@ flatpak_remote_state_ensure_metadata (FlatpakRemoteState *self,
...
       /* If the collection ID is NULL the metadata comes from the summary */
       if (self->metadata_fetch_error != NULL)
-        error_msg = g_strdup (self->metadata_fetch_error->message);
+        {
+          /* Fetch error can be IO_ERROR_CANCELLED in instances like quick searching of apps in software clients.
+           * Hence, avoid converting them to FLATPAK_ERROR from IO_ERROR domain so the clients can handle the
+           * cancellation appropriately. */
+          if (g_error_matches (self->metadata_fetch_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+            {
+              g_propagate_prefixed_error (error, g_steal_pointer (&self->metadata_fetch_error),
```
pwithnall: Same here. Not sure about stealing the `metadata_fetch_error`.

common/flatpak-dir.c
```diff
@@ -4551,8 +4582,18 @@ flatpak_dir_pull_extra_data (FlatpakDir          *self,
           g_autoptr(GError) my_error = NULL;
 
           if (!g_file_load_contents (extra_local_file, cancellable, &extra_local_contents, &extra_local_size, NULL, &my_error))
-            return flatpak_fail_error (error, FLATPAK_ERROR_INVALID_DATA, _("Failed to load local extra-data %s: %s"),
-                                       flatpak_file_get_path_cached (extra_local_file), my_error->message);
+            {
+              if  (g_error_matches (my_error, G_IO_ERROR, G_IO_ERROR_CANCELLED))
+                {
+                  g_propagate_prefixed_error (error, my_error, _("Failed to load local extra-data %s"),
```
pwithnall: Whereas here, `my_error` probably should be stolen, otherwise it’ll be double-freed.

pwithnall:
This looks OK to me now, but are you confident you’ve managed to hit all these code paths in testing?

uajain:
> This looks OK to me now, but are you confident you’ve managed to hit all these code paths in testing?

Yes, not in an individual/singleton testing but I have seen those errors over-time/multiple testing. With these changes, I haven't seen error-banner popping up anytime for "operation cancelled" with the current version of gnome-software downstream.

PR #162, opened 2019-06-11
fixup! profile/flatpak.sh: add export/bin dirs to PATH

mwleeds:
Use syntax compatible with OpenBSD ksh; see
https://github.com/flatpak/flatpak/pull/2947

https://phabricator.endlessm.com/T27001


mwleeds:
> Sure, but shouldn’t we also pull in the upstream patch to change it for `share_path`?

Sure, cherry-picked it

PR #165, opened 2019-06-19
flatpak-run: Also filter by content type on parental controls checks

andrunko:
This seems to work fine although I am not sure if that is the best way to get the desktop file/mimetypes there, please check and let me know if you have other suggestions.

Tested with:
```
andrunko@endless:~$ sudo malcontent-client set 1001 x-scheme-handler/http
App filter for user 1001 set
andrunko@endless:~$ DISPLAY=:0 /usr/bin/flatpak run --branch=stable --arch=x86_64 --command=eolie --file-forwarding org.gnome.Eolie http://google.com
error: Running app/org.gnome.Eolie/x86_64/stable is not allowed by the policy set by your administrator
```

After resetting the filter I can run `Eolie` again. Also running other apps that don't handle `x-scheme-handler/http` works as expected with and without the filter set.

https://phabricator.endlessm.com/T27009

### DIFF
--- a/common/flatpak-run.c
+++ b/common/flatpak-run.c
@@ -32,6 +32,7 @@
 #include <sys/personality.h>
 #include <grp.h>
 #include <unistd.h>
+#include <gio/gdesktopappinfo.h>
 #include <gio/gunixfdlist.h>
 #ifdef HAVE_DCONF
 #include <dconf/dconf.h>
@@ -3198,14 +3199,20 @@ regenerate_ld_cache (GPtrArray    *base_argv_array,
 
 static gboolean
 check_parental_controls (const char    *app_ref,
+                         FlatpakDeploy *app_deploy,
                          GCancellable  *cancellable,
                          GError       **error)
 {
+  g_auto(GStrv) app_ref_parts = NULL;
   g_autoptr(MctManager) manager = NULL;
   g_autoptr(MctAppFilter) app_filter = NULL;
   g_autoptr(GAsyncResult) app_filter_result = NULL;
   g_autoptr(GDBusConnection) system_bus = NULL;
   g_autoptr(GError) local_error = NULL;
+
+  app_ref_parts = flatpak_decompose_ref (app_ref, error);
+  if (app_ref_parts == NULL)
+    return FALSE;
 
   system_bus = g_bus_get_sync (G_BUS_TYPE_SYSTEM, NULL, error);
   if (system_bus == NULL)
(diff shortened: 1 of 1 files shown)


common/flatpak-run.c
```diff
@@ -3230,7 +3232,48 @@ check_parental_controls (const char    *app_ref,
                          _("Running %s is not allowed by the policy set by your administrator"),
                          app_ref);
 
-   return TRUE;
+  /* TODO: is there a better way to get the desktop file/mimetypes here? */
```
andrunko: Maybe using `g_desktop_app_info_new()` and then we could merge the filter checks and use `mct_app_filter_is_appinfo_allowed()` (which already checks by path, content type and appref) instead. But then other parts of the code seem to use this pattern to load the desktop file (nothing using `GAppInfo`/`GDesktopAppInfo`).

common/flatpak-run.c
```diff
@@ -3198,6 +3198,8 @@ regenerate_ld_cache (GPtrArray    *base_argv_array,
 
 static gboolean
 check_parental_controls (const char    *app_ref,
+                         const char    *app_id,
```
pwithnall: I know you’ve probably done it to avoid a `flatpak_decompose_ref()` call, but it seems a bit redundant to pass in both the `app_ref` and the `app_id`, given that the latter is one of the components of the former.

common/flatpak-run.c
pwithnall: I don’t know if flatpak will have added the right directory (`${deploy_path}/export/share/applications/`) to the set of paths checked by `g_desktop_app_info_new()` for desktop files by this point, which is probably why other bits of code don’t use `g_desktop_app_info_new()`.

You could use `g_desktop_app_info_new_from_filename()` and pass in the `desktop_file_path` you build below, though, which would save parsing the key file yourself.

common/flatpak-run.c
andrunko: Yes, I was on the fence here, I will update it, but maybe I dont need the `app_ref` anymore if I using `g_desktop_app_info_new_from_filename()` + `mct_app_filter_is_appinfo_allowed()`, will check.

common/flatpak-run.c
andrunko: Makes sense and I will try using `GDesktopAppInfo` here, but still wonder why other parts of the code seem to prefer parsing the key file directly (using `GKeyFile`).

pwithnall: Other parts of the existing upstream code, or our downstream modifications? If it’s the existing upstream code, then I guess that’s a question for `#flatpak` on freenode.

andrunko: Checking again, looks like the other places are using `GKeyFile` but to handle other key files (e.g. ostree repo config), not the desktop file. The only place that seems to be opening the desktop file with `GKeyFile` is `tests/testlibrary.c:check_desktop_file`.

common/flatpak-run.c
andrunko: Ended up keeping `app_ref` and using `flatpak_decompose_ref()` because to filter using the app info we need `app_deploy` (to build the desktop file path) which may be `NULL` here.

Although if the `app_deploy` is `NULL` it means that `app_ref` is for a `runtime`, which I believe we don't handle in malcontent/parental controls (although there is nothing prohibiting one to blacklist a runtime there with `mct_app_filter_builder_blacklist_flatpak_ref()`).

Please let me know what you think and if needed I can remove the separate check for whether the `app_ref` is allowed.

common/flatpak-run.c
andrunko: > Have you looked at how easy it would be to add unit tests for this? I’m wary we’re starting to build up more and more parental controls functionality without any unit tests.

I had a look and to add unit tests we'd need to add a mock accounts service extension for the AppFilter, etc - I'd suggest to handle this as a separate task, what do you think?

common/flatpak-run.c
pwithnall: LGTM.

common/flatpak-run.c
pwithnall: > I'd suggest to handle this as a separate task, what do you think?

Sure.

Use python-dbusmock for mocking up the accounts-service.

andrunko: Thanks for the review, added https://phabricator.endlessm.com/T27091 for the tests.

PR #168, opened 2019-07-05
T27185

uajain:
Flatpak keeps the maximum level of permission granted during an ongoing
transaction. Hence, it might happen that a runtime install can get
upgraded to "app.install" action in polkit.

Our downstream flatpak has amended polkit rules to enforce
parental controls restrictions. We check an "app.install" action +
if it is parental-control-ably allowed for the user. If so, the
user can install it.

However, a trivial case of app install starts with installing an app's
main ref followed by it's extension refs. An extension ref starts
with "runtime/". However, an extension ref will fail the parental
control polkit check because it does not have any appstream.

At this point, the extension ref installation has been upgraded
to "app.install" action because flatpak remembers maximum level
of permission it got during last op(the main ref installation).
Now, it checks if this extension ref is parentally allowed but
we know, it will fail. Hence, prompts an additional dialog.

We workaround this problem by detecting this scenario and setting
"parental-controls-is-appropriate" to "1" beforehand, hence bypassing
the check.

https://phabricator.endlessm.com/T27185

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -1802,6 +1802,7 @@ authorize_deploy_add_polkit_details (const gchar           *installation,
                                      const gchar           *origin,
                                      PolkitSubject         *subject,
                                      PolkitDetails         *details,
+                                     guint32                arg_flags,
                                      GError               **error)
 {
   g_autoptr(GError) local_error = NULL;
@@ -1869,6 +1870,21 @@ authorize_deploy_add_polkit_details (const gchar           *installation,
       polkit_details_insert (details,
                              "parental-controls-is-appropriate",
                              flatpak_appstream_check_rating (rating, app_filter) ? "1" : "0");
+    }
+  else if (g_str_has_prefix (ref, "runtime/") &&
+           (arg_flags & FLATPAK_HELPER_DEPLOY_FLAGS_APP_HINT) != 0)
+    {
+      /* This is the case where polkit action has been converted to "app-install" instead of
+       * "runtime-install" or "runtime-update". This happens because flatpak "keeps" max permissions
+       * to avoid multiple polkit dialogs.
+       *
+       * Given our "app-install" action is checked separately with parental-controls checks in polkit
+       * rules, this is a workaround to avoid an additional polkit dialog installation of a runtime or
+       * extension ref, if there was a prior "app-install" op executed in the same transaction */
+      g_debug ("No AppStream data found for runtime ref %s — disabling parental control "
+               "checks for this ref", ref);
+      polkit_details_insert (details,
+                             "parental-controls-is-appropriate", "1");
     }
   else
     {
@@ -2001,7 +2017,7 @@ flatpak_authorize_method_handler (GDBusInterfaceSkeleton *interface,
           if (!authorize_deploy_add_polkit_details (installation, system_bus,
                                                     get_sender_pid (invocation),
                                                     ref, origin,
-                                                    subject, details, &local_error))
+                                                    subject, details, flags, &local_error))
             {
               g_dbus_method_invocation_take_error (invocation, g_steal_pointer (&local_error));
               return FALSE;
@@ -2097,7 +2113,7 @@ flatpak_authorize_method_handler (GDBusInterfaceSkeleton *interface,
       if (!authorize_deploy_add_polkit_details (installation, system_bus,
                                                 get_sender_pid (invocation),
                                                 ref, remote,
-                                                subject, details, &local_error))
+                                                subject, details, flags, &local_error))
         {
           g_dbus_method_invocation_take_error (invocation, g_steal_pointer (&local_error));
           return FALSE;


system-helper/flatpak-system-helper.c
```diff
@@ -1870,6 +1871,21 @@ authorize_deploy_add_polkit_details (const gchar           *installation,
...
+       * "runtime-install" or "runtime-update". This happens because flatpak "keeps" max permissions
+       * to avoid multiple polkit dialogs.
+       *
+       * Given our "app-install" action is checked separately with parental-controls checks in polkit
+       * rules, this is a workaround to avoid an additional polkit dialog installation of a runtime or
+       * extension ref, if there was a prior "app-install" op executed in the same transaction */
+      g_debug ("No AppStream data found for runtime ref %s — disabling parental control "
+               "checks for this ref", ref);
+      polkit_details_insert (details,
+                             "parental-controls-is-appropriate", "1");
```
pwithnall: I’m a little concerned that there might be situations where this code is called in the *first* polkit check to be done. For example, if a new app needs a runtime which isn’t currently installed — the runtime will be installed before the app, but its polkit check will(?) be upgraded to `app-install`. If so, the whole lot would go through with `parental-controls-is-appropriate`, even though the app itself may not be appropriate.

Can that happen? I don’t know for sure, so please check the possibility.

If so, could we set `parental-controls-is-appropriate` here to the value of `flatpak_appstream_check_rating()` for the appstream data for the app, even if the polkit check is being done for the runtime or an extension? i.e. Can we look up the app associated with a particular runtime installation operation, and use that for the parental controls check?

uajain: > . For example, if a new app needs a runtime which isn’t currently installed — the runtime will be installed before the app, but its polkit check will(?) be upgraded to app-install.

Yes. This is the behaviour I can see/reproduce.

>  If so, the whole lot would go through with parental-controls-is-appropriate, even though the app itself may not be appropriate.

Does this mean, getting an (unnecessary) runtime deployed when we don't even know if the app's main ref will pass the polkit check or not? Or are you worried that runtime / extensions or related ref might bring in something inappropriate(how can we check that when they don't contain any appstream data) ?

> If so, could we set parental-controls-is-appropriate here to the value of flatpak_appstream_check_rating() for the appstream data for the app, even if the polkit check is being done for the runtime or an extension? i.e. Can we look up the app associated with a particular runtime installation operation, and use that for the parental controls check?

This is a good idea but demands technical feasibility. According to my understanding, each ref in the transaction is fetched and deployed one after the other. And the polkit check happens during the deployment.

To implement this, it might require to get all the refs involved in the transaction, do polkit checks on all of them beforehand and then proceed to fetch and deploy (I know these are very invasive changes).

pwithnall: > > If so, the whole lot would go through with parental-controls-is-appropriate, even though the app itself may not be appropriate.
> 
> Does this mean, getting an (unnecessary) runtime deployed when we don't even know if the app's main ref will pass the polkit check or not? Or are you worried that runtime / extensions or related ref might bring in something inappropriate(how can we check that when they don't contain any appstream data) ?

I’m worried that if app A needs runtime R, app A shouldn’t be installed due to being violent, and runtime R can always be installed (because it’s a runtime); then installing R first, upgrading its `runtime-install` polkit check to `app-install`, will allow A to be installed without passing the parental controls checks — because the parental controls checks for the whole transaction will have only been the ones for R.

> > If so, could we set parental-controls-is-appropriate here to the value of flatpak_appstream_check_rating() for the appstream data for the app, even if the polkit check is being done for the runtime or an extension? i.e. Can we look up the app associated with a particular runtime installation operation, and use that for the parental controls check?
> 
> This is a good idea but demands technical feasibility. According to my understanding, each ref in the transaction is fetched and deployed one after the other. And the polkit check happens during the deployment.
> 
> To implement this, it might require to get all the refs involved in the transaction, do polkit checks on all of them beforehand and then proceed to fetch and deploy (I know these are very invasive changes).

Yeah, that sounds painful. Given that it’s going to change upstream, maybe we could just do a hack here, and disable the code which upgrades `runtime-install` to `app-install`? That would possibly result in two polkit authentications for one flatpak transaction (one for the runtimes, and one for the apps), but would prevent the (likely, I think?) case of allowing an app to be installed when it should have failed parental controls checks.

However, please do test the scenario I described above and check it actually allows apps to be installed when they shouldn’t be. I might well have misunderstood how things work, and it might not be an issue at all. In which case, adding a hack to disable the `runtime-install` → `app-install` upgrade would be pointless.

uajain: > because the parental controls checks for the whole transaction will have only been the ones for R.

My understanding and quick testing with `polkit.log` says otherwise. 
Keeping permission-upgradation for the transaction aside, each ref will be checked though polkit rules (during it's deployment) . I can see each ref being checked via `polkit.log` action in a single transaction. There is no "one polkit check for the whole transaction" according to me. We should definitely double-check though.

uajain: Quick chat with mclasen on IRC:
```
[4:58:29 pm] <uajain> mclasen: Hey
[4:58:29 pm] <uajain> I have a question about polkit permission upgradation patch : http://github.com/flatpak/flatpak/commit/a2f57f64fda112938d86de032cd1f1e8d29ca376
[4:58:29 pm] <uajain> If I understand it correctly it simply takes the highest polkit permission it can i.e. app-install 
[4:58:29 pm] <uajain> A runtime can be installed via a polkit action app.install (if there is a transaction where a app needs a runtime install etc.)
[4:58:29 pm] <uajain> But this doesn't mean the polkit is checked only once per transaction. All the refs have to go through polkit check while their deployment happens..
[5:01:34 pm] <mclasen> yes
[5:02:11 pm] <mclasen> but we have a policy that allows polkit to keep auth, and we our actions have annotations to make e.g. app-install imply runtime-install
[5:03:28 pm] <mclasen> if you see multiple dialogs for a single transaction, you probably don't get the keep part of the policy
[5:03:38 pm] <mclasen> e.g. because you are remote and work via ssh
```

pwithnall: Do you have some results from your quick testing (i.e. `polkit.log` output) which you could share here for posterity?

uajain: > Do you have some results from your quick testing (i.e. `polkit.log` output) which you could share here for posterity?

Please bear with me with what I am about to post. I want to give out relevant details in a brief manner. I am not sure if that is sufficient enough.

* A flatpak install (client):
Screenshot from 2019-07-16 20-39-20

* `polkit.log` diff  : https://paste.gnome.org/pzfupbvet

* `flatpak-system-helper --verbose output` and `polkit.log` output (grepped out from journal, starting from line 69): https://paste.gnome.org/pfyxt482f

pwithnall: OK, looks good to me. :-)

PR #169, opened 2019-07-17
polkit: Admins should always be allowed to install to system repo

uajain:
In 42563c5273a, we applied parental controls to administrators too,
but forgot to take into account the polkit rule change that is needed
to retain the default behaviour for app-installation for admins.
Admins historically, were able to install apps without a polkit
challenge (i.e. without requiring additional authorization for
system repo installation).

This commit brings back the default behavior for admin users, being
able to install to system repo without a polkit challenge.

During next rebase, this commit can be squashed with 42563c5273a.

https://phabricator.endlessm.com/T26968

### DIFF
--- a/system-helper/org.freedesktop.Flatpak.rules.in
+++ b/system-helper/org.freedesktop.Flatpak.rules.in
@@ -8,7 +8,8 @@ polkit.addRule(function(action, subject) {
             return polkit.Result.YES;
     } else if (action.id == "org.freedesktop.Flatpak.app-install" &&
                subject.active == true && subject.local == true &&
-               action.lookup("parental-controls-repo-installation-allowed") === "1" &&
+               (subject.isInGroup("@privileged_group@") ||
+                action.lookup("parental-controls-repo-installation-allowed") === "1") &&
                action.lookup("parental-controls-is-appropriate") === "1") {
             return polkit.Result.YES;
     }


wjt:
retest this please

pwithnall:
bot, retest this please

pwithnall:
bot, retest this please

pwithnall:
If the tests fail this time, my guess is it’s because of https://github.com/endlessm/glib/pull/60.

andrunko:
> If the tests fail this time, my guess is it’s because of [endlessm/glib#60](https://github.com/endlessm/glib/pull/60).

Yes, that seems to be the case - in the worst case we can revert these changes.

I've spent a good time investigating this but could not figure it out what is leaking the `GDBusConnection` but at least I managed to nail it down a bit:
- Only 3 tests from `flatpak/tests/testlibrary.c` cause the leak:
   - `test_transaction_install_flatpakref`
   - `test_transaction_flatpakref_remote_creation`
   - `test_transaction_no_runtime`
- Disabling these 3 tests makes the test PASS/build succeed
- The problem seem to be related to invoking `flatpak_transaction_run` after invoking `flatpak_transaction_add_install_flatpakref`
- I initially thought it was some place leaking a `FlatpakDir` as it holds a `GDBusConnection` object until destruction, but after changing it to not hold the ref anymore the tests still leak
- Everywhere else using a `GDBusConnection|GDBusProxy` seems to either receive them as param (and not keep refs around) or use them together with `g_autoptr`
- I may have accidently found a few small issues with flatpak while investigating this
- See (very WIP) https://github.com/endlessm/flatpak/pull/170/commits/678de85ec7dd66a9bab042a2b54bc25c0045fe56 for my playground tests - [the build works with this commit](https://ci.endlessm-sf.com/job/obs-flatpak-pr/298/console)

I will continue on Monday if nobody beats me to it.

pwithnall:
Nice, some of those leak fixes are good catches which need to be submitted upstream ASAP.

Some debugging thoughts:
 * Does running the upstream tests under `valgrind` highlight anything?
 * Try running with `GOBJECT_DEBUG=objects` set in the environment, which will tell you all the `GObject` instances still alive at exit.
 * Could also try [this trick](https://tecnocode.co.uk/2010/07/12/reference-count-debugging-with-gdb/) with `gdb`, setting up the `g_object_ref` breakpoint only after your `// FIXME: no leak till here` point has been reached in a test. It’d be really time consuming though.

andrunko:
It turns out the changes to the tests using `flatpak_transaction_add_install_flatpakref` although it works in our jenkins builder it fails locally (still leaks), I am investigating more and trying @pwithnall suggestions - I had tried with valgrind already but didnt get too far, will check the other options.

andrunko:
> It turns out the changes to the tests using `flatpak_transaction_add_install_flatpakref` although it works in our jenkins builder it fails locally (still leaks), I am investigating more and trying @pwithnall suggestions - I had tried with valgrind already but didnt get too far, will check the other options.

Scratch that - the other tests were failing because I was missing the gvfs changes from https://obs-master.endlessm-sf.com/request/show/20418 and after applying them the only tests failing are the ones using `flatpak_transaction_add_install_flatpakref`.

andrunko:
The leak should be fixed - see https://phabricator.endlessm.com/T26968#738712.

Re-test here once the required changes land.

manuq:
Oh, this is why we had to start using `sudo` for system install of flatpaks at Hack?

wjt:
retest this please

andrunko:
retest this please

uajain:
> @uajain Changes look good but I would prefer if the commit msg used `fixup!` for [42563c5](https://github.com/endlessm/flatpak/commit/42563c5273a9c73c7ae2dd09ac860edcafb2a952) instead.
> 
> Leaving to you to decide and merge.

Merged about updating commit to be a `fixup!` commit text.
Thanks for the reviews and all the digging around / extra PRs.

PR #173, opened 2019-07-24
tests: Temporarily disable failing tests due to recent GLib changes

andrunko:
The issue was introduced with recent changes made to GLib[1]
where invoking g_test_dbus_down() will fail after a
timeout if the GDBusConnection object for the session bus leaks.

Lets disable these tests for now until the leaks are fixed.

[1] https://gitlab.gnome.org/GNOME/glib/merge_requests/963

https://phabricator.endlessm.com/T27345


wjt:
I think we should open a separate ticket (linked to T26968 and https://phabricator.endlessm.com/T25264) to track fixing these problems and reverting this patch.

andrunko:
> I think we should open a separate ticket (linked to T26968, https://phabricator.endlessm.com/T25264, and mentioned in the commit message) to track fixing these problems and reverting this patch.

Done - https://phabricator.endlessm.com/T27345 and updated commit msg.

PR #174, opened 2019-07-25
flatpak-installation: Handle an unconfigured remote when listing remotes

pwithnall:
This prevents a crash in flatpak_installation_list_remotes_by_type() if
the `FlatpakDir` can’t ensure it has a repo configured.

https://phabricator.endlessm.com/T27337

---

WIP: I am still testing this

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1197,7 +1197,8 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
 /* Find all USB and LAN repositories which share the same collection ID as
  * @remote_name, and add a #FlatpakRemote to @remotes for each of them. The caller
  * must initialise @remotes. Returns %TRUE without modifying @remotes if the
- * given remote doesn’t have a collection ID configured.
+ * given remote doesn’t have a collection ID configured or if the @dir doesn’t
+ * have a repo.
  *
  * FIXME: If this were async, the parallelisation could be handled in the caller. */
 static gboolean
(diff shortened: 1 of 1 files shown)


pwithnall:
Finished testing, and it seems to fix the crash

common/flatpak-installation.c
```diff
@@ -1227,8 +1229,12 @@ list_remotes_for_configured_remote (FlatpakInstallation *self,
       !types_filter[FLATPAK_REMOTE_TYPE_LAN])
     return TRUE;
 
+  repo = flatpak_dir_get_repo (dir);
+  if (repo == NULL)
+    return TRUE;
```
uajain: I am still trying to wrap my head around, in what cases a FlatpakDir's `repo` member can be `NULL` ?

But looking at how `list_remotes_for_configued_remote` is called, I think this `repo==NULL` and return check can possibly live in the `flatpak_installation_list_remotes_by_type`. (there is already a `repo != NULL` check there).

uajain: Tested the crash with https://paste.fedoraproject.org/paste/Ta~4EylNh0AJUyo9WcbCpg

Seems to fix it.

pwithnall: > I am still trying to wrap my head around, in what cases a FlatpakDir's `repo` member can be `NULL` ?

I’m not 100% sure about the semantics, but I think it’s basically when a `FlatpakDir` is created with a repository which doesn’t exist. For example, the first time when flatpak is started. Or if a user `FlatpakDir` is created for a user who doesn’t have `~/.local/share/flatpak`.

> But looking at how `list_remotes_for_configued_remote` is called, I think this `repo==NULL` and return check can possibly live in the `flatpak_installation_list_remotes_by_type`. (there is already a `repo != NULL` check there).

It could, but I think it’s better to keep the check nearer to the use of the `dir`, i.e. where I’ve put it in `list_remotes_for_configured_remote()`. Then we don’t get surprises if that function is called from somewhere else (and we forget to add a check there) in future.

pwithnall:
I’m going to go ahead and merge this so it doesn’t block testing; we can always rework the patch in a subsequent PR if you feel strongly about where the check goes (please say so if so).

PR #180, opened 2019-09-18
Rebase to flatpak upstream master (commit a641bb76) and set version to 1.4.99

andrunko:
Here are the details on the changes/patches:
- af9aea6e Update Indonesian translation
  - Dropped: translation
- c3df39a0 Update Ukrainian translation
  - Dropped: translation
- 5020790f Suppress debug output when completing
  - Dropped: upstream in 5020790f53f39d34410af669e1793c1ef370d43b
- 26ad9154 Add flatpak in docker seccomp profile
  - Dropped: upstream in 26ad9154c31a484062ffcb46a3818a39ae35a866
- ec2f4372 Checkout using the same mode of the repo
  - Refreshed
- 7eb813b4 build-finish: also export speedwagon files
  - Refreshed
- da748b0f build-finish: Add Endless Discovery Feed Providers
  - Refreshed
- 0bd9ae84 Allow defining an external location for the system installation and use it
  - Refreshed
- 2ad90bf1 Include the external installation directory in XDG_DATA_DIRS
  - Refreshed
- a7342555 flatpak-dir: skip rewriting search provider .ini files for Endless apps
  - Refreshed
- f594e23e common/dir: Change default of core.add-remotes-config-dir to false
  - Refreshed
- 74da5f37 profile/flatpak.sh: add export/bin dirs to PATH
  - Refreshed
- be4fbba2 flatpak-run: Add parental controls support for filtering apps
  - Dropped: Part of e1419b31dba0f45d712952f7adb89ac370f69c79 cherry-picked from pwith's upstream PR
- 7cf2cae5 flatpak-run: Ignore missing app filters for system users
  - Dropped: Part of e1419b31dba0f45d712952f7adb89ac370f69c79 cherry-picked from pwith's upstream PR
- 657f0b1f Revert "Move flatpak_dir_load_appstream_store"
  - Dropped: no longer needed with new parental controls
- f0be6dc2 dir: Support filtering app installs by user’s OARS settings
  - Dropped: Part of 230a8dd802cfbbe135b3231929f784b8641e6439 cherry-picked from pwith's upstream PR
- a129b3fd dir: Allow app filter to disallow app installation entirely
  - Dropped: Part of 230a8dd802cfbbe135b3231929f784b8641e6439 cherry-picked from pwith's upstream PR
- 1eb62556 system-helper: Move parental controls checks to system-helper
  - Dropped: Part of 230a8dd802cfbbe135b3231929f784b8641e6439 cherry-picked from pwith's upstream PR
[...]

PR #182, opened 2019-09-19
T25941 Fix offline g-s USB updates

mwleeds:
https://phabricator.endlessm.com/T25941


mwleeds:
This is now included in #180

PR #183, opened 2019-09-20
Backport changes to allow extensions to use extra-data (and lie about the version)

andrunko:
https://phabricator.endlessm.com/T27865


mwleeds:
I suppose the window of risk here is pretty small. Users will only have an incorrect version between when 3.6.4 is released and when 3.7.0 is released and they update, so probably only a few weeks.

mwleeds:
I'll leave the merge to @wjt since it's the stable branch

PR #184, opened 2019-09-24
common/dir: Skip parental controls checks when running from the eos-image-builder

andrunko:
The image builder doesn't have the system bus available which is required to
run the parental controls checks, so lets skip it if the
FLATPAK_SKIP_PARENTAL_CONTROLS_NO_SYSTEM_BUS env variable is set (set by the builder)
and system bus is not available.

https://phabricator.endlessm.com/T27896

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7653,6 +7653,7 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
 #ifdef HAVE_LIBMALCONTENT
   g_autoptr(GError) local_error = NULL;
   const char *on_session = g_getenv ("FLATPAK_SYSTEM_HELPER_ON_SESSION");
+  const char *skip_parental_controls_no_system_bus = g_getenv ("FLATPAK_SKIP_PARENTAL_CONTROLS_NO_SYSTEM_BUS");
   g_autoptr(GDBusConnection) dbus_connection = NULL;
   g_autoptr(MctManager) manager = NULL;
   g_autoptr(MctAppFilter) app_filter = NULL;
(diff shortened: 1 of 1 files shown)


mwleeds:
As Philip pointed out on Slack, if we just skip parental controls checks if the system bus isn't available, the user can circumvent parental controls by doing `DBUS_SYSTEM_BUS_ADDRESS=/dev/null flatpak install --user flathub org.doom.Doom` and with this approach they only have to do `DBUS_SYSTEM_BUS_ADDRESS=/dev/null FLATPAK_SKIP_PARENTAL_CONTROLS_NO_SYSTEM_BUS=1 flatpak install --user flathub org.doom.Doom`

So this doesn't seem any better in terms of security; it just has the advantage that we can distinguish between environments where we expect no system bus to exist, and ones where we do expect a system bus to exist.

andrunko:
@mwleeds Yes, I agree this doesn't add much re-security, although it makes it a tad more difficult to find out which vars to set. 

As this is a workaround to fix the builder, I'd suggest we go with this option and keep https://phabricator.endlessm.com/T27896 open to work on a proper fix, but please let us know if you have other suggestion.

andrunko:
Worth noting the @pwithnall also suggested this workaround on Slack.

mwleeds:
Okay. I suppose it's also not worse than the status quo where you can do `FLATPAK_SYSTEM_HELPER_ON_SESSION=1 flatpak install ...`

And my understanding is that before the recent rebase flatpak didn't do parental controls checks at all when running as root, and therefore didn't do them in image builds. So I suppose we expect people who work on image configurations to not put in apps with explicit content. That's probably fine; right now I don't think there's even a default parental controls limit set. Anyway I added a comment about that here: https://phabricator.endlessm.com/T26894

PR #186, opened 2019-09-26
installation: don't dereference possibly-NULL array

wjt:
If the installation contains 1 or more installed refs, but none of those
refs have a remote with a collection ID, then 'results' will be NULL but
'installed' will be non-NULL. Since
c29e686246086fa24a9cf9ba291d0e2f43420e06, 'results[0]' is used in this
situation – a NULL pointer dereference. There is an existing 'results !=
NULL' check inside the body of this loop, but this is too late.

Check whether 'results' is NULL before dereferencing it.

Submitted upstream as https://github.com/flatpak/flatpak/pull/3135.

https://phabricator.endlessm.com/T27912


mwleeds:
I cherry-picked the (only slightly different) upstream commit instead. Closing this.

PR #187, opened 2019-10-02
Fix more regressions

andrunko:
https://phabricator.endlessm.com/T27854

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7680,7 +7680,7 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
     {
       /* FIXME: Instead of skipping the parental controls check in the test
        * environment, make a mock service for it.
-       * https://phabricator.endlessm.com/T25340 */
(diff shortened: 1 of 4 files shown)


andrunko:
> Looks good, but I’d add something to the fixup commit messages to make it clear that they’re meant to be dropped in favour of the upstream versions when we next rebase.
> 
I think the fact that they are already fixup! commits kinda defeats this need? Also the other fixup commits we have in the same branch don't mention anything about dropping them in favour of the upstream versions and adding this info just in these new commits may add more confusion than help? If you still prefer to have this info let me know and I will update.

> I’d also add something to the “system-helper: Unconditionally allow 'app-install' polkit action” commit message to make it clear it’s a downstream commit which shouldn’t be dropped and shouldn’t be upstreamed.
Agreed, will update.

andrunko:
Updated only the commit msg for "system-helper: Unconditionally allow 'app-install' polkit action”, but please let me know what you think.

PR #191, opened 2019-10-21
T28028 flatpak 1.5.0 rebase (Debian packaging)

pwithnall:
**Do not merge** — it needs to be pushed manually.

Code changes in #192.

https://phabricator.endlessm.com/T28028

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+flatpak (1.5.0-1endless1) eos; urgency=medium
+
+  * Rebase against upstream 1.5.0 release (T28028)
+    - Drop various patches upstreamed between 1.4.3 and 1.5.0
+    - Add socat as a dependency of the tests
+
+ -- Philip Withnall <withnall@endlessm.com>  Mon, 21 Oct 2019 17:31:00 +0100
+
 flatpak (1.4.99-1endless1) eos; urgency=medium
 
   * Rebase against upstream master (commit a641bb76)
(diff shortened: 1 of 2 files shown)


pwithnall:
>     * There's Debian packaging of 1.5.0 [here](https://salsa.debian.org/debian/flatpak/commits/debian/experimental). We could rebase on that.

Hah, looks like Simon and I were working on this at basically the same time.

His packaging has [this commit](https://salsa.debian.org/debian/flatpak/commit/48210b941dc114acf279a078a8ad8f814b8db04c) which we don’t, but that’s because we already exposed that symbol in our `.symbols` file due to it being added downstream earlier. So that’s OK.

I don’t think it’s worth the time to rebase this PR on the upstream Debian stuff right now, as they don’t meaningfully differ, and I’d have to re-test everything. It can be done next time we need to rebase the packaging (1.6?).

>     * In the last rebase we dropped the old changelog entries. I'm fine with not doing that here but I thought it was worth mentioning.

I only drop changelog entries if I’m rebasing the Debian packaging. In this case, this PR is an incremental update to our existing packaging, rather than a rebase on `debian/experimental`.

Thanks for the review!

pwithnall:
Merged:
```
$ git push -u origin --force-with-lease --atomic T28028-rebase:master debian-T28028-rebase:debian-master Version_1.5.0_debian Version_1.5.0
Total 0 (delta 0), reused 0 (delta 0)
To github.com:endlessm/flatpak.git
   a2ee0cf0..44e0864c  debian-T28028-rebase -> debian-master
 + 42052f19...5bab2805 T28028-rebase -> master (forced update)
Branch 'debian-T28028-rebase' set up to track remote branch 'debian-master' from 'origin'.
Branch 'T28028-rebase' set up to track remote branch 'master' from 'origin'.
```

PR #195, opened 2019-11-02
utils: Allow chaining OstreeAsyncProgress when pushing GMainContext

ptomato:
It's a common idiom in this codebase to push a temporary GMainContext as
the thread default context in order to run an async operation as if it
were sync. If we are not expecting progress callbacks this isn't a
problem, but it becomes a problem if we pass in an OstreeAsyncProgress
object that was created under a different GMainContext. The reason for
this is that OstreeAsyncProgress creates an idle source and attaches it
to the thread default context, so if we are iterating a temporary
context then the OstreeAsyncProgress's context never gets iterated, and
so no progress signals are fired.

To fix this, we introduce flatpak_progress_chain() and a RAII helper
OstreeAsyncProgressChained which creates a new OstreeAsyncProgress under
the temporary GMainContext, but forwards all its state and updates to
the previous OstreeAsyncProgress's callbacks.

This is documented in a comment in the code as well.

All known instances of this problem in the existing code are fixed in
this commit.

This problem was solved conceptually by Philip Withnall, I only wrote
the code.

https://phabricator.endlessm.com/T26965

### DIFF
--- a/common/flatpak-utils-private.h
+++ b/common/flatpak-utils-private.h
@@ -812,6 +812,18 @@ typedef void (*FlatpakProgressCallback)(const char *status,
 OstreeAsyncProgress *flatpak_progress_new (FlatpakProgressCallback progress,
                                            gpointer                progress_data);
 
+OstreeAsyncProgress *flatpak_progress_chain (OstreeAsyncProgress *progress);
+
+static inline void
+flatpak_progress_unchain (OstreeAsyncProgress *chained_progress)
+{
+  if (chained_progress != NULL)
+    ostree_async_progress_finish (chained_progress);
+}
+
+typedef OstreeAsyncProgress FlatpakAsyncProgressChained;
+G_DEFINE_AUTOPTR_CLEANUP_FUNC (FlatpakAsyncProgressChained, flatpak_progress_unchain);
+
 void flatpak_log_dir_access (FlatpakDir *dir);
 
 gboolean flatpak_check_required_version (const char *ref,
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -6196,7 +6196,7 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
   guint64 bytes_transferred;
   guint64 fetched_delta_part_size;
   guint64 total_delta_part_size;
-  guint outstanding_extra_data;
+  guint64 outstanding_extra_data;
   guint64 total_extra_data_bytes;
   guint64 transferred_extra_data_bytes;
   guint64 total = 0;
@@ -6280,7 +6280,7 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
                              "requested", "u", &requested,
                              "start-time", "t", &start_time,
                              "status", "s", &status,
-                             "outstanding-extra-data", "u", &outstanding_extra_data,
+                             "outstanding-extra-data", "t", &outstanding_extra_data,
                              "total-extra-data-bytes", "t", &total_extra_data_bytes,
                              "transferred-extra-data-bytes", "t", &transferred_extra_data_bytes,
                              "downloading-extra-data", "u", &downloading_extra_data,
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -6217,11 +6217,13 @@ progress_cb (OstreeAsyncProgress *progress, gpointer user_data)
   guint64 total_transferred;
   g_autofree gchar *formatted_bytes_total_transferred = NULL;
   g_autoptr(GVariant) outstanding_fetchesv = NULL;
+  g_autoptr(GVariant) outstanding_extra_datav = NULL;
 
   /* We get some extra calls before we've really started due to the initialization of the
      extra data, so ignore those */
   outstanding_fetchesv = ostree_async_progress_get_variant (progress, "outstanding-fetches");
-  if (outstanding_fetchesv == NULL)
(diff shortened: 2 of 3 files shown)


ramcq:
http://giphygifs.s3.amazonaws.com/media/qIXVd1RoKGqlO/giphy.gif

common/flatpak-utils.c
```diff
@@ -6411,12 +6411,122 @@ flatpak_progress_new (FlatpakProgressCallback progress,
...
 }
 
+static void
+copy_ostree_progress_state (OstreeAsyncProgress *source,
+                            OstreeAsyncProgress *dest)
+{
+#define COPY_OSTREE_STATE(key) G_STMT_START {                          \
+  GVariant *variant = ostree_async_progress_get_variant (source, key); \
+  if (variant)                                                         \
+    ostree_async_progress_set_variant (dest, key, variant);            \
```
pwithnall: Calling `ostree_async_progress_set_variant()` multiple times will result in the progress update being made non-atomic. IIRC the existing code is careful to do all related progress updates in a single `ostree_async_progress_set()` call, so they remain grouped atomically.

We need an `OstreeAsyncProgress` method which returns the entire state of the progress,suitable for copying straight into another `OstreeAsyncProgress`. That might require API addition in upstream OSTree, and making this code be the fallback for if that’s not available.

common/flatpak-utils.c
```diff
@@ -6411,12 +6411,122 @@ flatpak_progress_new (FlatpakProgressCallback progress,
...
+ *     {
+ *       g_autoptr(GMainContextPopDefault) context = NULL;
+ *       g_autoptr(OstreeAsyncProgressChain) chained_progress = NULL;
+ *       context = flatpak_main_context_new_default ();
+ *       chained_progress = flatpak_progress_chain (progress);
+ *       ostree_some_async_op (chained_progress,
+ *                             some_callback_setting_some_flag, data);
+ *       while (wait_for_flag)
+ *         g_main_context_iteration (context, TRUE);
+ *     }
```
pwithnall: Might be worth mentioning in this comment that `OstreeAsyncProgressChained` is only safe to use when the two `GMainContext`s are in the same thread. Otherwise it will call the outer callback function from the wrong thread and cause thread safety issues.

common/flatpak-utils-private.h
```diff
@@ -812,6 +812,18 @@ typedef void (*FlatpakProgressCallback)(const char *status,
...
+OstreeAsyncProgress *flatpak_progress_chain (OstreeAsyncProgress *progress);
+
+static inline void
+flatpak_progress_unchain (OstreeAsyncProgress *chained_progress)
+{
+  if (chained_progress != NULL)
+    ostree_async_progress_finish (chained_progress);
+}
+
+typedef OstreeAsyncProgress OstreeAsyncProgressChained;
```
pwithnall: I think this should probably be `FlatpakAsyncProgressChained` (or something similar in the `Flatpak` namespace) so that you don’t wander into the `Ostree` namespace. That could cause issues if this functionality were to make its way into libostree in future, for example.

common/flatpak-utils.c
ptomato: I think I'll rewrite this to use ostree_async_progress_get()` and `ostree_async_progress_set()` in order to only lock/unlock twice, instead of 42 times, in any case. That would seem to be a better fallback than the current code.

ptomato: Ugh, tried that and it doesn't work. `ostree_async_progress_get()` fails an assertion if the key isn't present yet, whereas `ostree_async_progress_get_variant()` just returns null in that case. I can see why they did that, otherwise they'd have to define a default value for every key somehow. So this is the best fallback we can do, I think.

ptomato: Oh, that assertion fail is even occurring with the code already in this pull request, which previously worked.

I'm not sure why it's crashing but if I move the API into libostree then it works (and is better anyway) so I think we should do that. Unfortunately I think that means we wouldn't be able to write a reliable fallback in libflatpak.

ptomato:
Upstream pull request: https://github.com/flatpak/flatpak/pull/3211

pwithnall:
I had a thought about the `ostree_async_progress_get()` call aborting if the key didn’t exist: if you don’t want this to block on getting new API into OSTree (although I think that’s the right long-term fix), you could make additional changes in flatpak to ensure that all the known keys used in any `OstreeAsyncProgress` instance are pre-populated with a value when the `OstreeAsyncProgress` is constructed. Then you could use the old fallback code to copy all keys from one `OstreeAsyncProgress` to another and not need to block on new API in OSTree.

ptomato:
Updated to fix crash that I observed today. (Although I can't be 100% sure it's gone until tomorrow, I suppose...)

ptomato:
I've updated the pull request again with a fallback for the new libostree API. To be honest it's quite a hack and I'm not sure I'd recommend using it due to the potential for crashes due to inconsistent state. I'd prefer merging the libostree pull request downstream.

I experimented with a different fallback using `ostree_async_progress_get()` and `ostree_async_progress_set()`, and default values for every key to prevent `ostree_async_progress_get()` from crashing, but in the diminishing time I have to investigate this, I could not find a set of default values that would make both the flatpak CLI and gnome-software happy, always strange behaviour in one or the other. See https://gist.github.com/ptomato/e20d0a8592fa0d38fde1efc5fb9a55ac if you are curious.

ptomato:
bot, retest this please

(The ostree pull request has landed downstream)

ptomato:
Discussed in today's standup that we want to merge this downstream for 3.7.5 earlier rather than later so it can get plenty of testing.

PR #198, opened 2019-11-11
common/dir: set MIRROR flag consistently when pulling commits

ramcq:
The pull to fetch our commit metadata uses the MIRROR flag since
8cea78d, however this is not applied consistently in the other pulls
done in flatpak_dir_setup_extra_data and flatpak_dir_pull. The result
is that two refs are added to the OSTree transaction - one remote ref
and one collection ref, but only one is found in future by
flatpak_repo_resolve_rev so the other is shadowed and ultimately
leaked. Apply the same conditions and pass the MIRROR flag consistently
so that only one ref is created.

https://phabricator.endlessm.com/T28344


mwleeds:
Closing this since we will instead backport https://github.com/flatpak/flatpak/pull/3225

PR #200, opened 2019-11-20
T28344 Fix duplicated mirror refs issue

mwleeds:
This is a backport of the upstream PR https://github.com/flatpak/flatpak/pull/3225 with the exception of an unrelated commit. As noted in the commit message, there were minor merge conflicts.

https://phabricator.endlessm.com/T28344


common/flatpak-dir.c
```diff
@@ -14823,3 +14838,83 @@ static void
...
+  g_autoptr(GHashTable) collection_refs = NULL;  /* (element-type OstreeCollectionRef utf8) */
+  g_autoptr(GPtrArray) ignore_collections = g_ptr_array_new_with_free_func (g_free); /* (element-type utf8) */
+  g_auto(GStrv) remotes = NULL;
+  const char *repo_collection_id;
+  OstreeRepo *repo;
+  int i;
+
+  /* Generally a flatpak repo should not have its own collection ID set, but
+   * check just in case flatpak is being run on a server for some reason.
+   */
```
pwithnall: This approach to working out which mirror refs to delete feels wrong. If the problem we’re trying to solve is mirror refs being created which alias remote refs, then perhaps we should be deleting mirror refs whose collection ID is also represented by a remote?

Thinking about the situations we’d expect:
 * A server, which has a collection ID set, will have a load of refs in `refs/heads`, and we wouldn’t expect it to have any refs in `refs/remotes` or `refs/mirrors`*, because it’s just serving its own stuff. So all mirror refs could be deleted if `repo_collection_id != NULL`.
 * A client (which doesn’t do P2P) will have a load of `refs/remotes` set, no `refs/heads`, and no `refs/mirrors`. So all its `refs/mirrors` could be deleted.
 * A client, `C1`, which is doing P2P will have a load of `refs/remotes` set, no `refs/heads`, and a load of `refs/mirrors` which alias the `refs/remotes` so that other peers (`C2`) can pull refs without having to share the same remote name → collection ID mapping as `C1`. The `refs/mirrors` layout could be virtual (not on disk), and could be remapped to `refs/remotes` by `C1` when `C2` makes requests to it.

* It’s technically possible for a server to serve content from multiple collection IDs using `refs/mirrors`, but do any servers actually do that? Unless there’s a way to know exactly which `refs/mirrors` are now dangling, it might be safer to delete all `refs/mirrors`, although that makes me sad.

Or we could take the approach that server admins should be on top of their own disk space, so we only need to tidy up `refs/mirrors` if `repo_collection_id == NULL`. That would prevent us stomping on a use case of mirror refs in servers.

(When I say ‘all mirror refs could be deleted’, I mean the ones which are under flatpak’s control, like `app/*`.)

wjt: Consider the case where the app was pre-installed, so had a mirrors/, remotes/ and deploy/ ref. When uninstalled, the latter two were removed, but mirrors/ remains.

pwithnall: No, your point is relevant. I was considering that use case (but should have listed it explicitly). Your use case is a specialisation of the two client use cases above.

mwleeds: > It’s technically possible for a server to serve content from multiple collection IDs using refs/mirrors, but do any servers actually do that? Unless there’s a way to know exactly which refs/mirrors are now dangling, it might be safer to delete all refs/mirrors, although that makes me sad.

I guess it's unknowable whether this case exists in the wild, but if it does exist it would be because the server admin did `ostree pull --mirror`, and there's no known problem with that codepath, which would write only to `refs/mirrors/` and not `refs/remotes/`. (Presumably they're not using flatpak to install from the server to itself.)

So I don't see any reason why we should delete collection-refs that match repo_collection_id.

And deleting all flatpak-related mirror refs seems incorrect for another reason: that would mean deleting things from the `eos` remote which has `xa.disable=true`. So to know if that's safe we'd have to prove that the other clients of the repo, like `eos-updater` and `ostree`, don't create mirror refs in normal operation. And there's at least one ref, `ostree-metadata`, which is used by both flatpak and non-flatpak clients, so deleting it might interfere with the operation of e.g. eos-updater (I haven't checked but as written we don't delete `(com.endlessm.Os, ostree-metadata)`).

mwleeds: Oh, and there's another reason we need to ignore collection-refs that match `repo_collection_id`: `ostree_repo_list_collection_refs()` includes refs in `refs/heads/`, so otherwise we'd delete everything on a server and make someone very sad.

pwithnall: I think all this reasoning about which sets of refs to delete should go into a comment or commit message somewhere.

mwleeds: There are comments, but I can push a fixup to improve them

pwithnall: I think that would be helpful

mwleeds: Pushed a commit, if it looks good I can send it upstream.

PR #201, opened 2019-11-27
T28819: enable search based on the provider bus name, not the app ID

ramcq:
Currently we filter by app IDs to determine which search providers to enable, which means Hack and Endless Network knowledge apps are not searchable by default. Revert this patch, and instead check the BusName of search provider .ini files which are exported, and do not disable those where the search provider is from com.endlessm, so that all knowledge apps are searchable by default.

https://phabricator.endlessm.com/T28819

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -6541,7 +6541,12 @@ export_ini_file (int                 parent_fd,
     return FALSE;
 
   if (ini_type == INI_FILE_TYPE_SEARCH_PROVIDER)
-    g_key_file_set_boolean (keyfile, "Shell Search Provider", "DefaultDisabled", TRUE);
+    {
+      g_autofree gchar *bus_name = g_key_file_get_string (keyfile, "Shell Search Provider", "BusName", NULL);
+
+      if (bus_name == NULL || !g_str_has_prefix (bus_name, "com.endlessm."))
+        g_key_file_set_boolean (keyfile, "Shell Search Provider", "DefaultDisabled", TRUE);
+    }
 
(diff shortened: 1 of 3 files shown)


common/flatpak-dir.c
```diff
@@ -6541,7 +6541,12 @@ export_ini_file (int                 parent_fd,
     return FALSE;
 
   if (ini_type == INI_FILE_TYPE_SEARCH_PROVIDER)
-    g_key_file_set_boolean (keyfile, "Shell Search Provider", "DefaultDisabled", TRUE);
+    {
+      g_autofree gchar *bus_name = g_key_file_get_string (keyfile, "Shell Search Provider", "BusName", NULL);
+
+      if (bus_name == NULL || g_str_has_prefix (bus_name, "com.endlessm."))
```
wjt:
```suggestion
      if (bus_name == NULL || !g_str_has_prefix (bus_name, "com.endlessm."))
```

ramcq: yeah yeah yeah :P

wjt:
How's that?

mwleeds:
> LGTM from a correctness point of view. It's unfortunate that nothing is stopping a maliciously crafted flatpak from using a bus name that starts with "com.endlessm." for their search provider and thereby accessing search queries, but that attack is also possible before this patch, and I don't have a solution to propose. I suppose until we switch to Wayland the Flatpak sandboxing story isn't coherent anyway.

I suppose the solution would be to match the bus name exactly rather than match the prefix, but then this will break if we make a `SearchProviderV4`.

PR #203, opened 2019-12-05
installation: Cache remote states while querying installed refs

uajain:
Cache remote state so that each remote is only queried once
for getting its state.

Follow up from https://github.com/flatpak/flatpak/pull/3204#discussion_r349642324

This can be dropped in next rebase cycle.
Backported from https://github.com/flatpak/flatpak/pull/3259

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1019,6 +1019,7 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
   g_autoptr(GPtrArray) installed = NULL; /* (element-type FlatpakInstalledRef) */
   g_autoptr(GPtrArray) remotes = NULL; /* (element-type FlatpakRemote) */
   g_autoptr(GHashTable) remote_commits = NULL; /* (element-type utf8 utf8) */
+  g_autoptr(GHashTable) remote_states = NULL; /* (element-type utf8 FlatpakRemoteState) */
   int i, j;
   g_autoptr(FlatpakDir) dir = NULL;
   g_auto(OstreeRepoFinderResultv) results = NULL;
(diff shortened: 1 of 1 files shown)


common/flatpak-installation.c
```diff
@@ -1112,9 +1115,16 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
        * This makes sure that the ref (maybe an app or runtime) remains in usable
        * state and fixes itself through an update.
        */
-      state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, error);
+      state = g_hash_table_lookup (remote_states, remote_name);
       if (state == NULL)
-        continue;
+        {
+
```
uajain: Just noticed, this extra line got in upstream too, so I will fixup! that upstream-only.

common/flatpak-installation.c
```diff
@@ -1112,9 +1115,16 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
...
+          state = flatpak_dir_get_remote_state_optional (dir, remote_name, FALSE, cancellable, error);
```
mwleeds: We need to pass `NULL` instead of `error` here. Otherwise, we risk re-using an already set `GError` pointer. This bug existed before your patch, but still worth fixing.

mwleeds: Submitted as https://github.com/flatpak/flatpak/pull/3280

pwithnall: That has been merged upstream, so it probably makes sense to backport it downstream in this PR too.

pwithnall: LGTM other than that

pwithnall: Is this still relevant? Seems to have stalled.

common/flatpak-installation.c
uajain: Done [here](https://github.com/flatpak/flatpak/pull/3376/files)

uajain:
Apologies for dropping the ball on this. Backported @mwleeds's patch from upstream too and updated the PR.

PR #207, opened 2020-02-11
Debian rebase on 1.6.2-1

dbnicholson:
Rebase for the debian packaging on 1.6.2-1. Most everything was dropped. The list of commits we had:

pick 9ad1456c Use /var/endless-extra/flatpak as external directory for system installation
pick 1393ca45 Depend on malcontent >= 0.4.0
drop 54d5d096 Add new symbols to symbols file (upstream)
drop ef77c1f1 Remove debian/patches (upstream)
drop a2ee0cf0 Release 1.4.99-1endless1 (changelog only commit)
drop d052a3aa debian: Add socat as a dependency of flatpak-tests (upstream)
drop 44e0864c debian: Release 1.5.0-1endless1 (changelog only commit)

A diff just against 1.6.2-1 can be seen at https://github.com/endlessm/flatpak/compare/debian/1.6.2-1...T29218-1.6.1-rebase-debian.

A completed build is in https://obs-master.endlessm-sf.com/package/show/home:dbn:branches:eos:master:endless/flatpak

The master PR is in #206.

https://phabricator.endlessm.com/T29218

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,13 @@
+flatpak (1.6.2-1endless1) master; urgency=medium
+
+  * Rebase against Debian bullseye (T29218)
+  * Resync Endless packaging changes:
+    - Add malcontent build dependency
+    - Use /var/endless-extra/flatpak as external directory for system
+      installation
+
+ -- Dan Nicholson <nicholson@endlessm.com>  Tue, 18 Feb 2020 16:01:04 -0700
+
 flatpak (1.6.2-1) unstable; urgency=medium
 
   * New upstream stable release


mwleeds:
I changed the base branch from master to debian-master, which I think is what you intended?

dbnicholson:
Indeed, thanks.

dbnicholson:
I've rebased this now on 1.6.2-1 but kept the same branch to preserve the previous comments.

debian/changelog
```diff
@@ -1,3 +1,13 @@
+flatpak (1.6.2-1endless1) master; urgency=medium
+
+  * Rebase against upstream master (T29218)
```
mwleeds: s/master/1.6.2/ right?

mwleeds:
I don't understand how the build succeeded. `debian/control` requires xdg-desktop-portal 1.6 but we have 1.4.2 in OBS.

dbnicholson:
> I don't understand how the build succeeded. `debian/control` requires xdg-desktop-portal 1.6 but we have 1.4.2 in OBS.

For one it's a runtime dependency, and for two it's in Recommends. However, I should update it. Thanks for catching that. I missed it in my haste.

debian/changelog
dbnicholson: Ah, right. I think I just copied the same message from before.

dbnicholson:
> > I don't understand how the build succeeded. `debian/control` requires xdg-desktop-portal 1.6 but we have 1.4.2 in OBS.
> 
> For one it's a runtime dependency, and for two it's in Recommends. However, I should update it. Thanks for catching that. I missed it in my haste.

FWIW, the change for this is in https://github.com/endlessm/flatpak/commit/c64e154678d39d6c14bbb00da2d2951daf84c3ff.

debian/changelog
dbnicholson: I decided to say "Debian bullseye" since this changelog is referring to the packaging.

dbnicholson:
CR ready again.

debian/changelog
```diff
@@ -1,3 +1,13 @@
+flatpak (1.6.2-1endless1) master; urgency=medium
```
mwleeds: Historically we've had `eos` here where you have `master`. Is that intentional?

dbnicholson: Using `eos` is wrong, IMO. Our tools don't actually care, but what's listed there is the codename of distro that's being targeted. E.g., debian doesn't put `debian` there. They use `unstable` or `buster` (for example) if they were targeting a stable branch. Equivalently, if you were targeting an EOS stable branch instead of master, you'd want to put `eos3.7` there or something. I think I documented that in the wiki, but probably not.

mwleeds: Okay

PR #210, opened 2020-03-23
eos: Add metrics submission on parental controls blocking run/install

pwithnall:
UIs which present flatpaks should prevent the user from trying to run or
install apps which they aren’t allowed to because of their parental
controls policy.

Add metrics events from flatpak at the point where it prevents an app
being installed or run, so we have some statistics about possible places
where the UI is not sufficiently hiding disallowed flatpaks. All being
well, these metrics events should never be submitted.

This should not be upstreamed.

https://phabricator.endlessm.com/T28742

---

Debian changes in #211.


pwithnall:
bot, retest this please

common/flatpak-dir.c
```diff
@@ -7932,10 +7934,22 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
...
 
   if (!authorized)
-    return flatpak_fail_error (error, FLATPAK_ERROR_PERMISSION_DENIED,
-                               /* Translators: The placeholder is for an app ref. */
-                               _("Installing %s is not allowed by the policy set by your administrator"),
-                               ref);
+    {
+      /* Send a metrics event so we have an idea if there are any places which
+       * don’t block app installing from the UI. If everything is implemented
+       * correctly, this event should never be submitted.
```
mwleeds: Maybe add "(unless the user initiated the install from the command line)" to this comment?

common/flatpak-dir.c
```diff
@@ -7932,10 +7934,22 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
...
+       *
+       * See: https://phabricator.endlessm.com/T28741 */
+#define FLATPAK_PARENTAL_CONTROLS_INSTALL_EVENT "9d03daad-f1ed-41a8-bc5a-6b532c075832"
```
mwleeds: This needs to be added to the wiki, and similarly for the `run` event.

common/flatpak-dir.c
pwithnall: Good catch, done.

common/flatpak-dir.c
pwithnall: Both done. https://phabricator.endlessm.com/phriction/diff/545/?l=41&r=42

PR #214, opened 2020-04-11
Remove Transifex integration

wjt:
The main reason we added the Transifex setup for Flatpak was so that we could get translated strings ASAP for the parental controls work. That's all been upstreamed now and we don't really have any reason to diverge downstream.

We don't strictly need to revert these patches but it will make the situation clearer for whomever rebases Flatpak in the future.

https://phabricator.endlessm.com/T29427


wjt:
I submitted our outstanding downstream translations upstream: https://github.com/flatpak/flatpak/pull/3554

I didn't revert the commits which pulled these into our repo.

PR #216, opened 2020-04-21
dir: Don't include downgrades in update list

mwleeds:
Currently, if you are offline and a repo on a USB drive provides an
older version of something that you have installed, that will be
presented as an available update in `flatpak update` and GNOME Software
(which uses flatpak_installation_list_installed_refs_for_update()), and
the update will then fail due to being a downgrade. Fix the issue by
checking the timestamp on a remote commit and ignoring ones older than
the local commit. For list_installed_refs_for_update() this bug is a
recent regression from the commit "installation: Re-implement
list_installed_refs_for_update()", and for the update CLI command it
might be a more long-standing bug.

This patch can't go upstream since there the use of the libostree p2p
api has been removed.

https://phabricator.endlessm.com/T28651

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -3762,13 +3762,22 @@ flatpak_dir_resolve_free (FlatpakDirResolve *resolve)
 }
 
 static const char *
-find_latest_p2p_result (OstreeRepoFinderResult **results, OstreeCollectionRef *cr)
+find_latest_p2p_result (OstreeRepoFinderResult **results,
+                        OstreeCollectionRef     *cr,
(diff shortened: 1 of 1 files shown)


mwleeds:
I have checked that this patch makes `list_installed_refs_for_update()` no longer list downgrades, using the Python bindings.

mwleeds:
Adding a unit test sounds like a nice idea in theory but (a) it's [technically difficult](https://github.com/flatpak/flatpak/pull/1877) and (b) it would be made obselete by our next Flatpak rebase.

PR #218, opened 2020-05-07
dir: don't disable Kolibri's search provider

wjt:
The Flatpak upstream behaviour is to disable search providers by
default:

> Installing a search provider is a minor security issue, as it
> can see all that you search for in the desktop

Kolibri's search provider queries a local database, so does not pose a
security risk. We trust Learning Equality not to change this.

https://phabricator.endlessm.com/T28681


wjt:
This is a draft because:

1. I've not tested it at all
2. I am not sure whether the right approach is https://github.com/endlessm/eos-theme/pull/346, https://github.com/endlessm/flatpak/pull/218, or both

PR #224, opened 2020-06-16
dir: Fix an error path with unset GError

mwleeds:
This fixes a place where we omit passing a GError pointer and therefore
potentially return FALSE without setting the passed in error. This
causes flatpak_installation_list_installed_refs_for_update() to hit an
assertion failure, crashing libflatpak (and subsequently the App Center
when it's used).

The debug log provided by the user shows messages being printed from
flatpak_dir_prepare_resolve_p2p_refs_helper() for only a subset of the
refs being resolved, and then the log stops since the process crashes,
so that's how I know with near certainty that this patch fixes the
crash experienced by the user on the forum.

The upstream PR on the flatpak-1.6.x branch is here:
https://github.com/flatpak/flatpak/pull/3691

https://phabricator.endlessm.com/T30340

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4013,7 +4013,7 @@ flatpak_dir_prepare_resolve_p2p_refs_helper (FlatpakDir             *self,
       if (resolve->local_commit == NULL || latest_rev == NULL)
         continue;
 
-      if (!ostree_repo_load_commit (state->child_repo, resolve->local_commit, &commit_data, NULL, NULL))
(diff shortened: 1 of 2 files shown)


mwleeds:
force-pushed to pull in changes from https://github.com/flatpak/flatpak/pull/3692

mwleeds:
This has been merged upstream:

- "dir: Fix an error path with unset GError" was merged to `flatpak-1.6.x`
- "transaction: Add assertions to catch unset GError issues " was merged to `master`

PR #226, opened 2020-07-10
T29642 rebase on 1.8.x (packaging)

mwleeds:

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,13 @@
+flatpak (1.8.1-1endless1) master; urgency=medium
+
+  * Rebase against Debian bullseye (T29642)
+  * Resync Endless packaging changes:
+    - Add malcontent build dependency
+    - Use /var/endless-extra/flatpak as external directory for system
+      installation
+
+ -- Matthew Leeds <matthew.leeds@endlessm.com>  Fri, 10 Jul 2020 13:06:14 -0700
+
 flatpak (1.8.1-1) unstable; urgency=medium
 
   * New upstream stable release


mwleeds:
Excluding changelog entries, here's the diff between this branch and `debian-master`, with this branch as `flatpak2`:
```
diff -r debian/control ../flatpak2/debian/control
49a50
>  libzstd-dev,
53c54,55
<  python3 <!nocheck>,
---
>  python3:any,
>  python3-pyparsing,
diff -r debian/copyright ../flatpak2/debian/copyright
25c25
<  © 2018 Endless Mobile, Inc.
---
>  © 2018-2019 Endless Mobile, Inc.
diff -r debian/flatpak.install ../flatpak2/debian/flatpak.install
11a12
> usr/lib/sysusers.d
21a23
> usr/share/fish/vendor_completions.d
23c25
< usr/share/gdm/env.d
---
> usr/share/gdm/env.d/*                   usr/share/doc/flatpak/examples/etc/gdm3/env.d
Only in ../flatpak2/debian/: flatpak.README.Debian
diff -r debian/gbp.conf ../flatpak2/debian/gbp.conf
5c5
< upstream-branch = upstream/1.6.x
---
> upstream-branch = upstream/1.8.x
diff -r debian/libflatpak0.symbols ../flatpak2/debian/libflatpak0.symbols
16a17
>  flatpak_http_error_quark@Base 1.7.3
107d107
<  flatpak_oci_error_quark@Base 0.9.1
179a180
>  flatpak_transaction_add_sideload_repo@Base 1.7.0~git20200325
196a198
>  flatpak_transaction_operation_get_is_skipped@Base 1.7.3
200a203
>  flatpak_transaction_operation_get_related_to_ops@Base 1.7.3
221a225
>  flatpak_transaction_set_no_interaction@Base 1.7.3
Only in ../flatpak2/debian/: patches
diff -r debian/rules ../flatpak2/debian/rules
43a44
> 		--disable-auto-sideloading \
44a46
> 		--enable-gdm-env-file \
48a51
> 		--with-run-media-dir=/media \
```

mwleeds:
Actually we do need another patch here, to pass `--with-run-media-dir=/run/media` to configure

mwleeds:
I added a patch to pass `--enable-auto-sideloading` and `--with-run-media-dir=/run/media` to configure.

The only other change here compared to `debian-master` that looks important is there's `usr/lib/sysusers.d` in `flatpak.install` now, to let systemd create the `_flatpak` user for `flatpak-system-helper`. But [per the docs](https://www.freedesktop.org/software/systemd/man/sysusers.d.html#) systemd will only create a user defined in a sysusers.d file if it doesn't exist, so I think that should not conflict with our [existing base-passwd entry](https://github.com/endlessm/base-passwd/blob/master/passwd.master).

mwleeds:
Pushed a fix for `debian/flatpak.install` so it contains the sideload units.

pwithnall:
> ```
> diff -r debian/libflatpak0.symbols ../flatpak2/debian/libflatpak0.symbols
> 16a17
> >  flatpak_http_error_quark@Base 1.7.3
> 107d107
> <  flatpak_oci_error_quark@Base 0.9.1
> ```

Looks like neither of those symbols should ever have been exported. I’ve opened https://github.com/flatpak/flatpak/pull/3786 (which shouldn’t block this).

> The only other change here compared to `debian-master` that looks important is there's `usr/lib/sysusers.d` in `flatpak.install` now, to let systemd create the `_flatpak` user for `flatpak-system-helper`. But [per the docs](https://www.freedesktop.org/software/systemd/man/sysusers.d.html#) systemd will only create a user defined in a sysusers.d file if it doesn't exist, so I think that should not conflict with our [existing base-passwd entry](https://github.com/endlessm/base-passwd/blob/master/passwd.master).

Correct, `sysusers.d` entries are a no-op if the user is already listed in `base-passwd`, so it’s fine to leave them in the packaging.

PR #227, opened 2020-07-10
T29642 rebase on 1.8.x

mwleeds:

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -1106,7 +1106,9 @@ handle_configure (FlatpakSystemHelper   *object,
       return TRUE;
     }
 
-  if ((strcmp (arg_key, "languages") != 0) && (strcmp (arg_key, "extra-languages") != 0))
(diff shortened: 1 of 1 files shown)


mwleeds:
This rebase was pretty uneventful (only a couple merge conflicts docomented in the commit messages) so the review should focus mostly on whether any patches were inadvertently dropped and whether any upstream changes may cause problems.

mwleeds:
The final commit "Fetch ostree-metadata for backwards compat" is brand new so deserves a bit more careful review than the rest.

mwleeds:
Added three cherry-picks from upstream master that seem worth having

pwithnall:
“eos: Allow metrics to be optional ” could be squashed into “eos: Add metrics submission on parental controls blocking run/install” (see its commit message).

I can’t see any problems otherwise

mwleeds:
> “eos: Allow metrics to be optional ” could be squashed into “eos: Add metrics submission on parental controls blocking run/install” (see its commit message).

Did that and pulled six more commits from upstream master

PR #230, opened 2020-08-19
Properly validate configured collection IDs

mwleeds:
This commit is a follow-up for the commit "dir: Validate locally
configured collection IDs". Whereas in that commit we validate the
collection ID in flatpak_remote_state_fetch_commit_object(), here we do
it in _flatpak_dir_get_remote_state(), since fetch_commit_object() is
not called on the code path normally taken for a transaction (only if
try_resolve_op_from_metadata() fails and in other niche scenarios).

This should ensure that if someone erroneously sets a collection ID on a
remote, the error will be caught quickly. It also helps the eos-updater
unit tests pass.

https://phabricator.endlessm.com/T23150


mwleeds:
Upstream version of this is here: https://github.com/flatpak/flatpak/pull/3804

mwleeds:
Changed this per discussion [here](https://github.com/flatpak/flatpak/pull/3804#discussion_r474037941)

PR #231, opened 2020-08-21
T29642 Rebase on 1.8.2

mwleeds:
Here are the new commits that we didn't have from 1.8.1 or the subsequent merges:
```
7ee022bf Update pofiles for release
3edd38eb Bump version to 1.8.2 and update NEWS
aa739a18 oci-auth: Don't ask for authentication if anon auth fails with weird error
5c63fba4 run: Fix double free in the cups config file parser
d1d53eb7 run: Avoid hanging if xdg-dbus-proxy startup fails
d6083668 dir: Cache responses from localed and account service
f48a8232 Update variant-schema-compiler module
d33c322c exports: Try harder to export usrmerge symlinks as symlinks
a3b562ea portal: Make the SpawnStarted code more robust
f4c3ea5e Fix argument order of clone() for s390x in seccomp filter
```

### DIFF
--- a/doc/flatpak-remote-modify.xml
+++ b/doc/flatpak-remote-modify.xml
@@ -125,7 +125,8 @@
                 <listitem><para>
                     Mark the remote as not enumerated. This means the remote will
                     not be used to list applications, for instance in graphical
-                    installation tools.
+                    installation tools. It will also not be used for runtime dependency
+                    resolution (as with <option>--no-use-for-deps</option>).
                 </para></listitem>
             </varlistentry>
 
--- a/doc/flatpak-remote.xml
+++ b/doc/flatpak-remote.xml
@@ -110,7 +110,8 @@
                 </varlistentry>
                 <varlistentry>
                     <term><option>xa.noenumerate</option> (boolean)</term>
-                    <listitem><para>Whether this remote should be used when presenting available contents. Defaults to true.</para></listitem>
+                    <listitem><para>Whether this remote should be used when presenting available apps/runtimes,
+                    or when searching for a runtime dependency. Defaults to true.</para></listitem>
                 </varlistentry>
                 <varlistentry>
                     <term><option>xa.nodeps</option> (boolean)</term>


mwleeds:
The packaging side of this is on the branch `debian-T29642-1.8.2-rebase`

mwleeds:
> LGTM with some notes:
> 
>     * next rebase it would be easier if we could list the downstream changes and what was done with them (drop, pick, squash, etc)
> 

I can do that next time but in this case I just picked all the patches which weren't already part of 1.8.2 (ones we had backported).

>     * we should drop [1922b71](https://github.com/endlessm/flatpak/commit/1922b71dfafe9ac54aa746a999f067a5b77cd014) now that we are dropping speedwagon - https://phabricator.endlessm.com/T30293
> 

I don't think so. This flatpak package is used in the infrastructure too not just the desktop, so if we may need to build app updates that will go to older versions of Endless, we need to keep exporting speedwagon files.

>     * the debian branch removes a public symbol, needs checking if anything is using it and possibly rebuild reverse deps
> 

`flatpak_http_error_quark` was only made public in 1.7.3 so not long ago. I searched `FLATPAK_HTTP_ERROR` on the endlessm GitHub org and didn't find anything.

andrunko:
> > ```
> > * we should drop [1922b71](https://github.com/endlessm/flatpak/commit/1922b71dfafe9ac54aa746a999f067a5b77cd014) now that we are dropping speedwagon - https://phabricator.endlessm.com/T30293
> > ```
> 
> I don't think so. This flatpak package is used in the infrastructure too not just the desktop, so if we may need to build app updates that will go to older versions of Endless, we need to keep exporting speedwagon files.
> 

hmm, does that mean we should carry this patch forever? @wjt what do you think?

> > ```
> > * the debian branch removes a public symbol, needs checking if anything is using it and possibly rebuild reverse deps
> > ```
> 
> `flatpak_http_error_quark` was only made public in 1.7.3 so not long ago. I searched `FLATPAK_HTTP_ERROR` on the endlessm GitHub org and didn't find anything.

++

mwleeds:
> > > ```
> > > * we should drop [1922b71](https://github.com/endlessm/flatpak/commit/1922b71dfafe9ac54aa746a999f067a5b77cd014) now that we are dropping speedwagon - https://phabricator.endlessm.com/T30293
> > > ```
> > 
> > 
> > I don't think so. This flatpak package is used in the infrastructure too not just the desktop, so if we may need to build app updates that will go to older versions of Endless, we need to keep exporting speedwagon files.
> 
> hmm, does that mean we should carry this patch forever? @wjt what do you think?

As long as we may be building such app updates I think. Maybe @dbnicholson would have input?

wjt:
> I don't think so. This flatpak package is used in the infrastructure too not just the desktop, so if we may need to build app updates that will go to older versions of Endless, we need to keep exporting speedwagon files.

By definition all app updates we build will always be delivered to all versions of Endless, if users haven't upgraded.

Strictly speaking the infra package is not the same as the OS package – I think it's the `buster` branch.

The question is whether we are happy for apps to lose their splash screens on old Endless once we update the Flatpak version on our infrastructure. My money's on "yes" because we are happy to drop the splash screens entirely.

dbnicholson:
> > I don't think so. This flatpak package is used in the infrastructure too not just the desktop, so if we may need to build app updates that will go to older versions of Endless, we need to keep exporting speedwagon files.
> 
> By definition all app updates we build will always be delivered to all versions of Endless, if users haven't upgraded.
> 
> Strictly speaking the infra package is not the same as the OS package – I think it's the `buster` branch.
> 
> The question is whether we are happy for apps to lose their splash screens on old Endless once we update the Flatpak version on our infrastructure. My money's on "yes" because we are happy to drop the splash screens entirely.

Yes, it's the [buster](https://github.com/endlessm/flatpak/commits/buster) branch. I think it's fine to drop these patches from master, and the next time we rebase for the infrastructure we'll just have to keep them. I don't know the best place to put them, but for lack of a better place I [added a comment](https://github.com/endlessm/flatpak/commit/9d814e324f4e5e65b86ce1abab904381e65dd518#commitcomment-41788891) on this commit.

dbnicholson:
That said, WRT to speedwagon splash images, do any apps we ship actually have them anymore? Hmm, looks like at least the encyclopedia released this week has one. Presumably SOMA supplies them and will continue to for some time. So, yes, I think we would want to keep this on the infrastructure side until we're not shipping apps that provide splash images or no longer support OS releases that use them.

mwleeds:
Should we also drop "build-finish: Add Endless Discovery Feed Providers"? Looks like that feature is being dropped: https://phabricator.endlessm.com/T30294

andrunko:
> Should we also drop "build-finish: Add Endless Discovery Feed Providers"? Looks like that feature is being dropped: https://phabricator.endlessm.com/T30294

Good catch, I missed this one - I think we should drop it also.

mwleeds:
Dropped those two commits

ramcq:
> > By definition all app updates we build will always be delivered to all versions of Endless, if users haven't upgraded.
> > Strictly speaking the infra package is not the same as the OS package – I think it's the `buster` branch.
> > The question is whether we are happy for apps to lose their splash screens on old Endless once we update the Flatpak version on our infrastructure. My money's on "yes" because we are happy to drop the splash screens entirely.
> 
> Yes, it's the [buster](https://github.com/endlessm/flatpak/commits/buster) branch. I think it's fine to drop these patches from master, and the next time we rebase for the infrastructure we'll just have to keep them. I don't know the best place to put them, but for lack of a better place I [added a comment](https://github.com/endlessm/flatpak/commit/9d814e324f4e5e65b86ce1abab904381e65dd518#commitcomment-41788891) on this commit.

Why keep the cost/complexity of the delta in our infrastructure? Most users who are online upgrade to the tip of our release branch, so most anyone who can get new apps also gets the new OSes unless they are stuck on 3.3.x. The apps team is going to be asked to start moving our apps to Flathub pretty soon, seeing as soon we won't actually need SOMA to build knowledge apps any more...

PR #237, opened 2020-10-15
Make listing refs work for sideload repos

mwleeds:
This commit makes
flatpak_installation_list_remote_refs_sync() work for local URIs
pointing at sideload repos, e.g.
"file:///media/mwleeds/flatpaks/.ostree/repo" where a command "flatpak
create-usb /media/mwleeds/flatpaks ..." was previously run. This is
important because the Endless fork of gnome-software uses that API to
list apps on a USB drive. The API worked before Flatpak 1.7.1 so making
it work again basically involves reverting parts of
336a127f5515f69c7d5c6aa7943076035c59bf84.

This commit also makes the remote-ls command work for file:// URIs to
sideload repos, which is useful behavior for debugging, and is also the
behavior mentioned in this blog post:
https://blogs.gnome.org/mclasen/2018/08/26/about-flatpak-installations/

https://phabricator.endlessm.com/T30368

### DIFF
--- a/app/flatpak-builtins-remote-ls.c
+++ b/app/flatpak-builtins-remote-ls.c
@@ -164,7 +164,8 @@ ls_remote (GHashTable *refs_hash, const char **arches, const char *app_runtime, 
       g_hash_table_iter_init (&iter, refs);
       while (g_hash_table_iter_next (&iter, &key, &value))
         {
-          char *ref = key;
+          FlatpakCollectionRef *coll_ref = key;
+          char *ref = coll_ref->ref_name;
           char *partial_ref;
           const char *slash = strchr (ref, '/');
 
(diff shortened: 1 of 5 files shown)


mwleeds:
Blocked on the upstream version: https://github.com/flatpak/flatpak/pull/3909

mwleeds:
Will have to also look into the unit test failure. Locally the `testlibrary` test fails on testlibrary.c:1845 with the message "Aborted due to failure"

mwleeds:
Upstream went with [a different solution](https://github.com/flatpak/flatpak/pull/3940). We have [a downstream ticket](https://phabricator.endlessm.com/T31121) about checking that it works for us.

PR #238, opened 2020-10-20
profile: Don't add current directory to PATH

dbnicholson:
If `new_dirs` is empty because the flatpak exported `bin` directories
are already in `PATH`, then `PATH` will contain a trailing `:`. An empty
element in `PATH` means that the current directory should be searched.
While some users may want this behavior, it was not the intention of the
change here to prefix `PATH` with the flatpak exported `bin`
directories. The change here is to only prefix `:` when `new_dirs` is
not empty. This is similar to the `XDG_DATA_DIRS` mangling which only
suffixes `:` if `new_dirs` is not empty.

This can be squashed into dacf0a884ea9d52e763d2f5cb1d720c6fdf7be2f on
the next rebase.

I was surprised to find that programs were being searched for in the current directory and found this to be the culprit. I don't know if the commit title should be in `fixup!` format or not for future rebasing.

### DIFF
--- a/profile/flatpak.sh
+++ b/profile/flatpak.sh
@@ -46,5 +46,5 @@ if command -v flatpak > /dev/null; then
     )
 
     export PATH
-    PATH="${PATH:-/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games}:${new_dirs:+${new_dirs}}"
+    PATH="${PATH:-/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games}${new_dirs:+:${new_dirs}}"
 fi


wjt:
Good catch. @mwleeds will know whether the `fixup!` convention is preferred here.

mwleeds:
> Good catch. @mwleeds will know whether the `fixup!` convention is preferred here.

Yeah I think a `fixup! ...` commit summary would be good.

mwleeds (approved):
Seems like questionable behavior on Bash's part to let the current directory be added implicitly rather than explicitly, but the patch looks good.

I updated the commit message to be in `fixup!` form.

mwleeds:
(Merged despite CI failure because it was just the check for `fixup!` commits, which was intentional in this case)

dbnicholson:
> Seems like questionable behavior on Bash's part to let the current directory be added implicitly rather than explicitly, but the patch looks good.

FWIW, from bash(1):

> A zero-length (null) directory name in the value of PATH indicates the current directory.  A null directory  name  may appear as two adjacent colons, or as an initial or trailing colon.

dash has the same documented feature, which makes me think it's in POSIX. Yep - from the definition for `PATH` in https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html:

> A zero-length prefix is a legacy feature that indicates the current working directory. It appears as two adjacent colons ( "::" ), as an initial colon preceding the rest of the list, or as a trailing colon following the rest of the list. A strictly conforming application shall use an actual pathname (such as .) to represent the current working directory in PATH .

Yay for legacy features!

PR #239, opened 2020-10-20
Upgrade infra buster build to 1.8.2

dbnicholson:
1.8.2 is needed on the build infrastructure now. This is just master as of 465db077e5a0a506fd47250e57b79beaf63c19a6 with 2 additional commits:

* build-finish: Add Endless Discovery Feed Providers
* build-finish: also export speedwagon files

Support for these features has been removed from EOS master and possibly all apps. However, since potentially old apps will be built and distributed to previous versions of Endless, I decided to keep them a bit longer. You can see the delta relative to master here:

https://github.com/endlessm/flatpak/compare/master...T30984-buster-flatpak-1.8.2

https://phabricator.endlessm.com/T30984

PR #240, opened 2020-10-20
Upgrade infra buster packaging to 1.8.2

dbnicholson:
1.8.2 is needed on the infrastructure now. This is debian-master as of d2b51829c261f4f4a9b11a65c69bc493570a9914 with 2 additional commits:

* Drop eosmetrics build dependency for infrastructure
* Drop malcontent build dependency for infrastructure

Neither of those are available on buster and the features they provide aren't used during builds. The changes relative to debian-master can be seen here:

https://github.com/endlessm/flatpak/compare/debian-master...T30984-debian-buster-flatpak-1.8.2

https://phabricator.endlessm.com/T30984

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+flatpak (1.8.2-1endless1buster1) buster; urgency=medium
+
+  * Rebuild for infrastructure (T30984)
+    - Remove malcontent build dependency
+    - Remove eosmetrics build dependency
+
+ -- Dan Nicholson <dbn@endlessos.org>  Tue, 20 Oct 2020 13:18:45 -0600
+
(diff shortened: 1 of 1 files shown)


dbnicholson:
The buster branch changes are in #239. Please just review and I'll force push this when ready.

mwleeds (commented):
On the existing `debian-buster` branch there's a `Version_1.6.2_1endless2buster1` tag whereas here there's only the version tag used on `debian-master`. Looks good otherwise, feel free to merge if that doesn't matter.

dbnicholson:
I didn't push the tag yet so it didn't get picked up by jenkins in case I had to modify the branch. I can push it if you'd like, though.

PR #248, opened 2020-11-12
installation: Include renamed apps when listing updates

mwleeds:
Currently if an app has been renamed and you run "flatpak update" you
will be prompted to decide whether to install the new name of the app
(and uninstall the existing one). However, in such a situation
flatpak_installation_list_installed_refs_for_update() does not return
the app, and therefore gnome-software doesn't see any updates available
(except in the case where the installed commit is older than the commit
with the eol-rebase metadata).

So include end-of-life-rebased apps in the returned list. Patches to
make gnome-software update such apps properly are forthcoming.

I checked that the included unit test only passes with this patch to
libflatpak.

(cherry picked from commit d406cf0ed7cf3d7293d59d281392f3059026f7de)

(There were some minor merge conflicts during cherry-pick, due to the
upstream commit d7716f41ac1c082e8d4758a819512df7fb1c9efd)

https://phabricator.endlessm.com/T31068

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -979,6 +979,26 @@ flatpak_installation_list_installed_refs_by_kind (FlatpakInstallation *self,
 }
 
 static gboolean
+end_of_lifed_with_rebase (FlatpakTransaction *transaction,
+                          const char         *remote,
+                          const char         *ref,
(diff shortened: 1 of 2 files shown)


mwleeds:
This was merged upstream.

PR #252, opened 2020-11-19
Don't auto-install sources for extensions

mwleeds:
There was a bug in the extension point matcher which made it
install `org.gnome.Totem.Videosite.YouTubeDl.Sources` (in addition to
`org.gnome.Totem.Videosite.YouTubeDl`) for the `org.gnome.Totem.Videosite`
extension.

We just need to make sure we only match the extension prefix if there
is a single element in the extension name following the extension
name (i.e. '.YouTubeDl', not '.YouTubeDl.Sources').

This fixes https://github.com/flatpak/flatpak/issues/3973

(cherry picked from commit f650c303e3edff8127ca24fb631b09fcebb36084)

https://phabricator.endlessm.com/T31110

### DIFF
--- a/common/flatpak-utils.c
+++ b/common/flatpak-utils.c
@@ -2720,6 +2720,8 @@ flatpak_summary_match_subrefs (GVariant   *summary_v,
           VarRefMapEntryRef entry = var_ref_map_get_at (ref_map, i);
           const char *cur;
           const char *id_start;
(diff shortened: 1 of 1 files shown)


mwleeds:
This is a clean cherry-pick from upstream.

andrunko (approved):
LGTM.

PR #253, opened 2020-11-20
dir: Fix spurious permission denied errors

mwleeds:
Currently this error is happening on the eos3.9 dev branch of Endless
OS:

Nov 19 12:05:55 endless eos-updater-flatpak-installer[464]:
eos-updater-flatpak-installer: Couldn’t apply some flatpak update
actions for this boot: Error opening directory
'/root/.local/share/flatpak/app': Permission denied

The reason is that we have a check in flatpak_dir_list_unused_refs() to
check if the per-user installation exists and only try to list the refs
there if so, but the existence check falsely succeeds due to the systemd
sandboxing on eos-updater-flatpak-installer.service, and
flatpak_dir_list_refs() then fails in find_used_refs(). Specifically the
ProtectHome=yes systemd service option makes /root inaccessible, and due
to a bug in GLib[1] this makes g_file_query_exists() falsely return TRUE
for any paths under /root.

So, check for not found and permission denied errors rather than doing
an existence check, as we should be doing anyway to avoid a
time-of-check/time-of-use race, as explained in the
g_file_query_exists() documentation.

[1] https://gitlab.gnome.org/GNOME/glib/-/issues/1237

https://phabricator.endlessm.com/T31131

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -15179,14 +15179,21 @@ flatpak_dir_list_unused_refs (FlatpakDir         *self,
    */
   if (!flatpak_dir_is_user (self))
     {
-      g_autoptr(GFile) user_base_dir = flatpak_get_user_base_dir_location ();
-      if (g_file_query_exists (user_base_dir, cancellable))
-        {
-          g_autoptr(FlatpakDir) user_dir = flatpak_dir_get_user ();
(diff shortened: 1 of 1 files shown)


mwleeds:
Upstream PR: https://github.com/flatpak/flatpak/pull/3982

pwithnall (approved):
[Merged upstream](https://github.com/flatpak/flatpak/pull/3982#event-4019691790)

PR #254, opened 2020-12-09
flatpak-installation: Add default dep sources to list installed refs

pwithnall:
It seems that all `FlatpakTransaction`s add the default dep sources, so
the internal transaction used to list installed refs for updates should
do the same.

This fixes a bug where
`flatpak_installation_list_installed_refs_for_update()` would return an
error saying “The application x requires the runtime y which was not
found” if the app was installed in the user repo, the runtime was
installed in the system repo, and no remote was configured (or one was
configured `xa.noenumerate=true`) in the user repo to provide the
runtime. If a remote was configured, the error wouldn’t be returned, but
the app would be spuriously listed for an update as its runtime couldn’t
be found.

Upstream: https://github.com/flatpak/flatpak/pull/3994
https://phabricator.endlessm.com/T30233

### DIFF
--- a/common/flatpak-installation.c
+++ b/common/flatpak-installation.c
@@ -1090,6 +1090,9 @@ flatpak_installation_list_installed_refs_for_update (FlatpakInstallation *self,
   if (transaction == NULL)
     return NULL;
 
+  /* CLI transactions set this. */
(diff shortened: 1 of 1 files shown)


pwithnall:
Trivial backport of https://github.com/flatpak/flatpak/pull/3994, already merged upstream

PR #262, opened 2021-03-30
system-helper: Set interactivity on the FlatpakDir from D-Bus call flags

pwithnall:
The system helper was already correctly using the `NO_INTERACTION` flag
in the D-Bus call flags to determine whether polkit calls from
`flatpak_authorize_method_handler()` should allow interactivity.

However, the system helper was not setting the no-interaction property
on the `FlatpakDir` used in the subsequent operation. When parental
controls are enabled, this sometimes results in polkit allowing
interaction when prompting for the `override-parental-controls` action,
even if the D-Bus call which activated the system helper specified the
`NO_INTERACTION` flag.

https://phabricator.endlessm.com/T31351

---

Simple backport of https://github.com/flatpak/flatpak/pull/4199, with some minor conflict resolution done on calls to `dir_get_system()`.

### DIFF
--- a/system-helper/flatpak-system-helper.c
+++ b/system-helper/flatpak-system-helper.c
@@ -224,6 +224,7 @@ schedule_idle_callback (void)
 static FlatpakDir *
 dir_get_system (const char *installation,
                 pid_t       source_pid,
+                gboolean    no_interaction,
                 GError    **error)
 {
   FlatpakDir *system = NULL;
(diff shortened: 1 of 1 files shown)


pwithnall:
Merged upstream, so I’ll merge it here too: https://github.com/flatpak/flatpak/pull/4199#event-4555282346

PR #267, opened 2021-06-01
transaction: Don’t abort on uninstall if deploy metadata is missing

pwithnall:
If the deploy metadata is missing for the locale runtime of an app which
is being uninstalled, flatpak will currently abort on an assertion
failure.

Prevent that abort, at the expense of not uninstalling the locale
runtime. A better fix could be found for this in future.

Fixes: #3943
Upstream: https://github.com/flatpak/flatpak/pull/4292

https://phabricator.endlessm.com/T32010

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -3145,6 +3145,11 @@ resolve_ops (FlatpakTransaction *self,
           /* We resolve to the deployed metadata, because we need it to uninstall related ops */
 
           metadata_bytes = load_deployed_metadata (self, op->ref, &checksum, NULL);
+          if (metadata_bytes == NULL)
+            {
+              op->skip = TRUE;
+              continue;
+            }
           mark_op_resolved (op, checksum, NULL, metadata_bytes, NULL);
           continue;
         }


pwithnall:
Untested, since it’s a trivial backport which has been reviewed upstream already.

wjt:
I saw these CI failures because of some Mono package being discussed on the Flatpak Matrix channel.

Do we want to knock these actions out of our fork of this repo? At best they are redundant, but unrelated build failures like this are confusing, and the "Block Autosquash Commits" one will be actively inconvenient since this is a repo that we do periodically rebase and squash.

dbnicholson:
> I saw these CI failures because of some Mono package being discussed on the Flatpak Matrix channel.
> 
> Do we want to knock these actions out of our fork of this repo? At best they are redundant, but unrelated build failures like this are confusing, and the "Block Autosquash Commits" one will be actively inconvenient since this is a repo that we do periodically rebase and squash.

I think we should disable them and then (should we want) [enable specific actions](https://docs.github.com/en/github/administering-a-repository/managing-repository-settings/disabling-or-limiting-github-actions-for-a-repository#allowing-specific-actions-to-run). As you say, these are at best redundant since we already have a CI flow that builds our specific environment and we really don't need to care about things like clang or old glib or valgrind.

dbnicholson:
Anyways, I think the change is correct and I'm going to merge this, but we can discuss the GitHub Actions part further in https://phabricator.endlessm.com/T32039.

dbnicholson (approved):

PR #268, opened 2021-06-11
Backport various fixes needed to fix eos-updater unit tests

pwithnall:
The eos-updater unit tests (specifically, the tests which use flatpak) build various test flatpaks and install them using eos-updater-flatpak-installer in various situations. In order to test that there are no assumptions about the architecture in use, all these flatpaks and tests are done for a fictitious `arch` architecture (rather than, say `x86_64`). This used to work fine, but was broken in libflatpak by the transition to the new summary/subsummary architecture.

It was fixed upstream in https://github.com/flatpak/flatpak/issues/4252, but the fix is not in our version of flatpak (1.10.2). This backports it, plus some other fixes from master which look useful (all memory corruption/leak fixes). One of those leak fixes, 756b9eae, is needed to avoid cherry-pick conflicts when backporting https://github.com/flatpak/flatpak/pull/4256.

All of the cherry-picks completed without conflicts. The only complication is that this adds the new public symbol `FLATPAK_QUERY_FLAGS_ALL_ARCHES` but it’s an enum member so shouldn’t throw Debian’s symbol versioning.

Helps https://phabricator.endlessm.com/T31918, but is not directly related.

### DIFF
--- a/app/flatpak-builtins-build-init.c
+++ b/app/flatpak-builtins-build-init.c
@@ -100,7 +100,7 @@ ensure_extensions (FlatpakDeploy *src_deploy, const char *default_branch,
                   g_autoptr(FlatpakDir) src_dir = NULL;
                   g_autoptr(GFile) deploy = NULL;
                   g_autoptr(GBytes) deploy_data = NULL;
-                  const char **subpaths;
+                  g_autofree const char **subpaths = NULL;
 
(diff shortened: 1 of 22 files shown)


dbnicholson (approved):
Seems reasonable. The early changes all seem legit and the multiarch stuff is presumably pretty well tested.

PR #270, opened 2021-09-28
flatpak-dir: Fix parental controls checks for root

pwithnall:
These checks were broken in commit d762a2f, as the commit failed to
consider the fact that `flatpak_dir_check_parental_controls()` is run
both in the `flatpak` CLI process run by the user, but also in the
`flatpak-system-helper` process which always runs as root, and which
handles any installations done on the system repository.

As a result, parental controls were not working for the system
repository.

Fix that by limiting the scope of the check to only pass if running
without the system helper. flatpak calls from root never go through the
system helper.

Fixes: #4418

https://phabricator.endlessm.com/T32545

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -7980,7 +7980,6 @@ flatpak_dir_check_parental_controls (FlatpakDir    *self,
 #ifdef USE_SYSTEM_HELPER
   g_autoptr(GError) local_error = NULL;
   const char *on_session = g_getenv ("FLATPAK_SYSTEM_HELPER_ON_SESSION");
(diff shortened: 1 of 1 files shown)


dbnicholson (approved):
Very nice! Probably the upstream "root doesn't need parental controls" change is what we should have done in the first place (and upstreamed it).

pwithnall:
> Probably the upstream "root doesn't need parental controls" change is what we should have done in the first place (and upstreamed it).

Yeah, agreed. I think at the time I was reluctant to treat `root` specially and was hoping for a very pure and beautiful implementation. Reality then slowly collided with that illusion and here we are. :grin:

PR #275, opened 2022-02-08
Rebase on 1.12.4

dbnicholson:
This is a rebase on 1.12.4. Here is our current commit queue on top of 1.10.7:

- [x] 72872bea Checkout using the same mode of the repo
- [x] feca2fd8 common/dir: Change default of core.add-remotes-config-dir to false

In theory we could drop these now that we have migrated users to a
separate system repo with the standard mode. However, that happens post
upgrade to EOS 4, so it's not guaranteed to have happened yet. I believe
we would need to add a check in the EOS4->EOS5 checkpoint and ensure
EOS4 flatpak's doesn't drop these commits.

- [x] 073d4836 profile/flatpak.sh: add export/bin dirs to PATH
- [x] 4ea429fe system-helper: Unconditionally allow 'app-install' polkit action
- [x] 30fa2edd dir: don't disable search providers from com.endlessm
- [x] 92d21470 eos: Add metrics submission on parental controls blocking run/install
- [x] e9a7448e Fetch ostree-metadata for backwards compat

I believe we want to continue with these downstream changes. The 2nd one
seems like it could be fixed upstream but would need work. The last one
may not be needed anymore.

- [ ] 24ba153a tests: Skip test using system inst for now

I tested without this in OBS and it succeeded. I believe this is fixed
by a926776cf4fcacc2d096e10bf1578a0e7c626cc7.

- [ ] c5cd2510 flatpak-dir: Make parental controls polkit query interactive iff dir is
- [ ] c8168e11 tests: web-server: fix an issue with the web server cache
- [ ] eed2514e system-helper: Set interactivity on the FlatpakDir from D-Bus call flags
- [ ] c165234d dir: Avoid a crash when looking up summary for a ref without an arch
- [ ] d09d7cb7 tests: Fix test-summaries.sh when running tests as root
- [ ] 567784c8 transaction: Don’t abort on uninstall if deploy metadata is missing
- [ ] 2badd5fc remote-list: Check for unlikely libostree error
- [ ] 9b0f6716 utils: Fix use-after-free and logic error
- [ ] 3f0685b9 Fix memory errors w/ use of var_arrayofstring_to_strv()
- [ ] 846d6451 context: Fix leak of env var key/value pair
- [ ] 11444980 common: Fix several memory leaks
- [ ] 0c7f75f8 remote-info: Fix printing of branch
- [ ] 0bdf271a remote-info: Fix memory leak of branch
- [ ] 2442fdf2 dir: Fix a GString leak
- [ ] 85cf1bfe list: Add a g_assert() for a coverity warning
- [ ] c086cd42 utils: Add missing check for empty string collection id
- [ ] 16b7e4b5 Add FLATPAK_QUERY_FLAGS_ALL_ARCHES for list_remote_refs()
- [ ] 1d0c985c Transaction: Ensure we download the subsummary for the arch of added refs
- [ ] 0a774b0a builtin-utils: Use shared helper
- [ ] 1b6abfe9 Include more architectures when listing installed refs.
- [ ] 6b78eb04 CLI: List arch if it is not the primary arch (even if unique)
- [ ] 7ba7bf47 flatpak-transaction: Squash error propagation and fix return value
- [ ] 16b88145 flatpak-transaction: Fix finding runtime with non-default remote+arch

Upstream backports.

The post-rebase commits can be seen in https://github.com/endlessm/flatpak/compare/Version_1.12.4...T32944-rebase-1.12.4.

https://phabricator.endlessm.com/T32944

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -160,6 +160,11 @@ static gboolean flatpak_dir_cleanup_remote_for_url_change (FlatpakDir   *self,
                                                            const char   *url,
                                                            GCancellable *cancellable,
                                                            GError      **error);
+static gboolean _flatpak_dir_fetch_remote_state_metadata_branch (FlatpakDir         *self,
+                                                                 FlatpakRemoteState *state,
+                                                                 gboolean            only_cached,
+                                                                 GCancellable       *cancellable,
+                                                                 GError            **error);
 
 static gboolean flatpak_dir_lookup_remote_filter (FlatpakDir *self,
                                                   const char *name,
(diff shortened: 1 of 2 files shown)


starnight (approved):
LGTM!

dbnicholson:
> LGTM!

Thanks for the review!

PR #276, opened 2022-02-08
Rebase on debian 1.12.4-1~bpo11+1

dbnicholson:
This rebases the packaging on 1.12.4-1~bpo11+1 from bullseye-backports. The major difference between the testing and backports version is relaxing the libseccomp dependency, which we need unless we import testing's libseccomp. Here is our commit queue on top of our previous 1.10.7-0+deb11u1.

- [x] 72f717dd debian: Add dependency on eos-metrics-0-dev for downstream metrics
- [x] bf95b519 Enable sideloading with correct media dir

The 1st is needed for metrics and the 2nd is probably still desired so
that sideloading from USB works automatically.

The new queue can be seen in https://github.com/endlessm/flatpak/compare/debian/1.12.4-1_bpo11+1...T32944-rebase-1.12.4-debian.

https://phabricator.endlessm.com/T32944

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+flatpak (1.12.4-1~bpo11+1endless1) master; urgency=medium
+
+  * Rebase on Debian Bullseye backports 1.12.4 release (T32944)
+    - Add eos-metrics dependency
+    - Enable USB sideloading from /run/media
+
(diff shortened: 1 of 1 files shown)


starnight (approved):
LGTM!

PR #278, opened 2022-04-28
Upgrade bullseye to 1.10.7 debian packaging

dbnicholson:
This is debian-eos4.0 + the following infra commits:

- [] 981c2145 Drop malcontent build dependency for infrastructure

Since bullseye has malcontent, we can just let it build against it.

- [x] 056631e5 Drop eosmetrics build dependency for infrastructure

But we still don't have eosmetrics in the infrastructure.

See https://github.com/endlessm/flatpak/compare/debian-eos4.0...T33441-bullseye-1.10.7-debian for the github view.

https://phabricator.endlessm.com/T33441

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,10 @@
+flatpak (1.10.7-0endless1+deb11u1) bullseye; urgency=medium
+
+  * Rebuild for infrastructure (T33441)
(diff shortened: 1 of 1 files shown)


jprvita (approved):
I can't merge through the UI due to conflicts, feel free to force-push

PR #279, opened 2022-05-17
Rebase on Flatpak 1.12.7

wjt:
No merge conflicts.

https://phabricator.endlessm.com/T33491

PR #280, opened 2022-05-17
Rebase on Flatpak 1.12.7 (Debian packaging)

wjt:
No merge conflicts, except the changelog entry patch.

Code counterpart in #279 

https://phabricator.endlessm.com/T33491

### DIFF
--- a/debian/changelog
+++ b/debian/changelog
@@ -1,3 +1,11 @@
+flatpak (1.12.7-1~bpo11+1endless1) master; urgency=medium
+
+  * Rebase on Debian Bullseye backports 1.12.7 release (T33491)
+    - Add eos-metrics dependency
+    - Enable USB sideloading from /run/media
+
+ -- Will Thompson <wjt@endlessos.org>  Tue, 17 May 2022 16:13:57 +0100
+
 flatpak (1.12.7-1~bpo11+1) bullseye-backports; urgency=medium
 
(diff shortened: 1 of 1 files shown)


wjt:
I'm not wild about how this packaging branch is based on the Debian branch that has the full code in it as well. But I followed suite with `git rebase -i debian/1.12.7-1_bpo11+1` and I have to say it was much less painful and error-prone than the other workflow of manually unpacking the source tarball then reapplying our changes is.

Maybe :exploding_head: there should be just … one branch … based on the Debian package?

dbnicholson:
> I'm not wild about how this packaging branch is based on the Debian branch that has the full code in it as well. But I followed suite with `git rebase -i debian/1.12.7-1_bpo11+1` and I have to say it was much less painful and error-prone than the other workflow of manually unpacking the source tarball then reapplying our changes is.
> 
> Maybe exploding_head there should be just … one branch … based on the Debian package?

I was just thinking about that. In ye olden days, it was preferred that the main branch was a clean rebase on upstream so that you could see exactly what patches we had outstanding. In those days salsa git branches also didn't exist. Maybe we feel differently these days. I would be fine with that, but I think we should bring that discussion to the whole OS team list. Rebasing vs merging probably comes into play, too. I also have an idea to make the debian handling a little nicer WRT tags and changelog entries.

dbnicholson (approved):

PR #281, opened 2022-11-01
[eos4.0] flatpak-dir: Clean up temp deploy dir on failure of flatpak_dir_deploy()

pwithnall:
This already happens for installs due to the cleanup path in `flatpak_dir_deploy_install()`, but it doesn’t happen for other calls to `flatpak_dir_deploy()`. Notably, during updates of already installed apps.

Specifically, this means that if an app update is cancelled due to being blocked by a parental controls policy, the temp deploy dir for that app (such as
`~/.local/share/flatpak/app/com.corp.App/x86_64/stable/.somehex-XXXXXX`) will be leaked. It will never be automatically cleaned up, as it’s not in `/var/tmp` either.

Fix that by using `glnx_mkdtempat()` to create a scoped temporary directory.

(cherry picked from commit 85a83a06f954080bd1fbba5f07b465955e34040c with minor conflicts)
Upstream: https://github.com/flatpak/flatpak/pull/5159

https://phabricator.endlessm.com/T33923

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4378,6 +4378,70 @@ flatpak_dir_remove_appstream (FlatpakDir   *self,
   return TRUE;
 }
 
+#define SECS_PER_MINUTE (60)
+#define SECS_PER_HOUR   (60 * SECS_PER_MINUTE)
+#define SECS_PER_DAY    (24 * SECS_PER_HOUR)
+
+/* Like the function above, this looks for old temporary directories created by
+ * previous versions of flatpak_dir_deploy().
+ * These are all directories starting with a dot. Such directories can be from a
+ * concurrent deploy, so we only remove directories older than a day to avoid
+ * races.
+*/
+static void
+remove_old_deploy_tmpdirs (GFile *dir)
+{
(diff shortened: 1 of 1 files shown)


dbnicholson (commented):
LGTM, but I think we should hold and see how flatpak/flatpak#5164 shakes out. I think the cleanup is going to be as important as avoiding the problem.

pwithnall:
Updated to use the upstream cleanup function from https://github.com/flatpak/flatpak/pull/5164. I only pulled in one of the two commits there, as the flatpak in eos4.0 doesn’t have `remove_old_appstream_tmpdirs()`, which one of the commits trivially modifies. It’s not needed for anything other than debugging.

The second commit didn‘t cherry-pick cleanly, so might need a bit of review. The conflict was easy to resolve though.

Ready for review.

pwithnall:
bot, rebuild this please

dbnicholson (approved):
Looks good. I see how the conflict would have worked with the lack of the `remove_old_appstream_tmpdirs` function (and the missing `SECS_*` macros).

PR #282, opened 2022-11-01
flatpak-dir: Clean up temp deploy dir on failure of flatpak_dir_deploy()

pwithnall:
This already happens for installs due to the cleanup path in `flatpak_dir_deploy_install()`, but it doesn’t happen for other calls to `flatpak_dir_deploy()`. Notably, during updates of already installed apps.

Specifically, this means that if an app update is cancelled due to being blocked by a parental controls policy, the temp deploy dir for that app (such as
`~/.local/share/flatpak/app/com.corp.App/x86_64/stable/.somehex-XXXXXX`) will be leaked. It will never be automatically cleaned up, as it’s not in `/var/tmp` either.

Fix that by using `glnx_mkdtempat()` to create a scoped temporary directory.

(cherry picked from commit ce1829a703f2bc2e7868fd314ddefb63fbf7dce1) Upstream: https://github.com/flatpak/flatpak/pull/5158

https://phabricator.endlessm.com/T33923

### DIFF
--- a/common/flatpak-dir.c
+++ b/common/flatpak-dir.c
@@ -4489,6 +4489,66 @@ remove_old_appstream_tmpdirs (GFile *dir)
     }
 }
 
+/* Like the function above, this looks for old temporary directories created by
+ * previous versions of flatpak_dir_deploy().
+ * These are all directories starting with a dot. Such directories can be from a
+ * concurrent deploy, so we only remove directories older than a day to avoid
+ * races.
+*/
+static void
+remove_old_deploy_tmpdirs (GFile *dir)
+{
+  g_auto(GLnxDirFdIterator) dir_iter = { 0 };
+  time_t now = time (NULL);
+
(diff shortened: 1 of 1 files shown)


pwithnall:
Packaging changes here: https://github.com/endlessm/flatpak/pull/283

pwithnall:
It looks like cleaning up leaked directories will be a little safer to do upstream, so I’ve opened https://github.com/flatpak/flatpak/pull/5164. That should be pulled into this PR once it’s reviewed upstream. Marking this as a draft until then.

dbnicholson (commented):
This looks completely reasonable to me. IMO, flatpak/flatpak#5164 is also sane and preferable to the tmpfiles snippet. I'd be willing to roll with it if the upstream review stalls for too long.

dbnicholson (approved):

PR #285, opened 2023-01-18
Fix fetching ostree metadata commit

dbnicholson:
The aim here is to fix regressions in 438e8115 so that flatpak continues to fetch the `ostree-metadata` commit when the remote has a collection ID. Unfortunately, the current test in `test-sideload.sh` is skipped in our build configurations since it requires `bwrap` to work. This contains a new test script, `test-ostree-metadata.sh` that doesn't require any special privileges.

In order to perform the test, the `ostree-metadata` commit needs to be created and included in the summary. Previously this was done with a hack to run `ostree summary -u` before `flatpak build-update-repo`. That would still work for the compat ostree format summary, but it does not affect the new flatpak indexed summaries. This would also be used in our infrastructure when it updates to a newer flatpak. See https://phabricator.endlessm.com/T33445.

I'm marking this as a draft as I don't know if we want to bother with this anymore. The reason to support this is to allow USB's created with current flatpak to be consumed by flatpak on EOS <= 3.8. If that's not desired, then we can revert 438e8115 and move on with our lives.

https://phabricator.endlessm.com/T34320

### DIFF
--- a/app/flatpak-builtins-repo.c
+++ b/app/flatpak-builtins-repo.c
@@ -227,6 +227,22 @@ print_branches_for_subsummary (FlatpakTablePrinter *printer,
           const char *metadata;
           const char *eol;
 
+          /* Endless-specific: Print the ostree-metadata ref if it's
+           * included in the summary.
+           */
+          if (g_strcmp0 (ref, OSTREE_REPO_METADATA_REF) == 0)
+            {
+              g_autofree char *size = g_format_size (0);
+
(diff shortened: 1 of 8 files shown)


pwithnall:
This is superseded by https://github.com/endlessm/flatpak/pull/286, right?

dbnicholson:
Indeed, I'll close this. I'm not going to delete the branch just yet, though, as I may use the first 2 commits on the infrastructure to keep supporting old clients.

PR #290, opened 2023-02-16
transaction: Ignore uninstall operations for no deploy

dbnicholson:
If `no_deploy` has been set to `TRUE` in a transaction, then the intention is that no changes will be made to the installed flatpaks. Currently that's not the case for explicitly or implicitly added uninstall operations. That's particularly bad for eol-rebase flatpaks since they old version will be automatically removed without the new version being installed. To address this, prevent uninstall operations from being added for no deploy transactions.

Closes: #5172

https://phabricator.endlessm.com/T34071

This is taken from flatpak/flatpak#5289. It's almost a clean cherry pick of the one commit, but there have been some minor changes in `tests/test-repo.sh` that caused some conflicts.

### DIFF
--- a/common/flatpak-transaction.c
+++ b/common/flatpak-transaction.c
@@ -2112,6 +2112,13 @@ add_related (FlatpakTransaction          *self,
           if (!rel->delete)
             continue;
 
+          if (priv->no_deploy)
+            {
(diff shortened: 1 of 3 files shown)


pwithnall (approved):
:ship:

PR #298, opened 2024-03-12
Update to 1.14.5 plus 2 patches

wjt:
This is a simple `git merge flatpak-1.14.x`.

Packaging:

- [ ] #299 

https://phabricator.endlessm.com/T35258


wjt:
retest this please

dbnicholson:
retest this please

dbnicholson:
The reason for the previous failed build is because of some stale doc build stamp files in the directory causing the HTML build to be skipped. It seems like this has happened more lately and it was stumping me as to why. Looking at 3.9 build log reminded me why. In the dirty `jenkins-run-chroot` days, the git repo was cloned into the chroot to ensure you didn't end up with a bunch of root owned files in the Jenkins managed repo. That means you'd always be starting from a clean checkout. I think we need to add some `git clean` action now that this is all running on the original repo.

wjt:
Merged https://github.com/endlessm/eos-build/pull/1512/ to do that; waiting for the templates to be deployed before retesting.

dbnicholson (approved):
Looks like a winner!

